PR #1, opened 2017-01-24
Node control tool

azheregelya:
Small python script to call system-dependent scrips to upgrade Sovrin node
And the same work for Windows


azheregelya:
Outdated

PR #9, opened 2017-01-31
Additional fixes for upgrade procedure

azheregelya:
Try to download required package first and then install it from cache


azheregelya:
Cancelled

PR #16, opened 2017-02-01
Corrected the mode of opening a file for write via csv.writer.

spivachuk:
This fixed the following tests which had previously failed on Windows 10:
- test_update_log,
- testNodeDetectsUpgradeDone,
- testSendNodeUpgradeToAllNodes,
- testNodeSchedulesUpgradeAfterRestart.

### DIFF
--- a/sovrin_node/server/upgrade_log.py
+++ b/sovrin_node/server/upgrade_log.py
@@ -58,7 +58,7 @@ class UpgradeLog:
         now = datetime.utcnow()
         event = (now, type, when, version)
 
-        with open(self.__filePath, mode="a+") as file:
+        with open(self.__filePath, mode="a+", newline="") as file:
             writer = csv.writer(file, delimiter=self.__delimiter)
             writer.writerow(event)
         self.__items.append(event)


sovrin_node/server/upgrade_log.py
```diff
@@ -58,7 +58,7 @@ def __append(self, type, when, version) -> None:
         now = datetime.utcnow()
         event = (now, type, when, version)
 
-        with open(self.__filePath, mode="a+") as file:
+        with open(self.__filePath, mode="a+", newline="") as file:
```
lovesh: Why do you want to return the line endings? We are not using it anyway.

spivachuk: Before the fix:
Dialect.lineterminator is '\r\n' for csv.writer by default (dialect='excel') while newline parameter of open function is None by default.
From the doc of open function:
"When writing output to the stream, if newline is None, any '\n' characters written are translated to the system default line separator, os.linesep. If newline is '' or '\n', no translation takes place."
So '\n' character itself was translated to '\r\n' on Windows. All this produced the sequences '\r\r\n' for newlines. When reading, such a sequence was interpreted as two new lines, so empty lines appeared. Attempts to parse these empty lines as events failed.

There are posts about this issue:
http://stackoverflow.com/questions/3191528/csv-in-python-adding-an-extra-carriage-return
http://stackoverflow.com/questions/1170214/python-2-csv-writer-produces-wrong-line-terminator-on-windows

If we use newline='' for open function when writing then '\n' is written "as is". Thus with Dialect.lineterminator being '\r\n', the new lines are written as '\r\n' and later are correctly interpreted when reading.

In Python 3 documentation the recommended ways for opening files for csv.reader and csv.writer are with newline=''. Please see:
https://docs.python.org/3/library/csv.html

lovesh: @spivachuk Thanks for correcting me.

PR #18, opened 2017-02-03
Upgrade failed timeout 

mzk-vct:
Support of timeout after which started upgrade will be declared as failed


mzk-vct:
Must be merged to master instead of dev. Closing.

PR #42, opened 2017-02-20
Clean data script

mzk-vct:
Script for removing data and configuration from node


scripts/clear_node.py
```diff
@@ -0,0 +1,77 @@
...
+        for dir in TARGET_DIRS:
+            if os.path.exists(dir):
+                shutil.rmtree(dir)
+        return
+
+    files_to_keep = [re.compile(pattern) for pattern in WHITE_LIST]
+    def isOk(path):
+         return any(pattern.match(path) for pattern in files_to_keep)
+
+    for dir in TARGET_DIRS:
```
lovesh: `dir` is a reserved keyword, avoid using it

PR #52, opened 2017-03-07
rename SPONSOR to TRUST_ANCHOR

ashcherbakov:
Updating CLI and transaction names/parameters to reflect latest terminology:
- renamed SPONSOR to TRUST_ANCHOR
- use encoded values for roles (instead of names)
- supported possible backward-compatible rename of roles in future


sovrin_node/test/test_suspension.py
```diff
@@ -1,12 +1,11 @@
 import pytest
 
 from plenum.common.eventually import eventually
+from plenum.common.txn import TRUSTEE, TGB, STEWARD, TRUST_ANCHOR
```
dhh1128: Note my hesitation about the use of the TGB role.

dhh1128:
I am requesting some minor tweaks (see my comments in the reviews). I left my code review in the Approved state despite this request, because it would be acceptable to me to merge as-is, and to do the tweaks after the fact in a separate PR. Please read my notes and make an intelligent choice. Back to you, @ashcherbakov .

sovrin_node/test/test_suspension.py
ashcherbakov: Discussed with Lovesh:
it was discussed with Jason sometime, and approved by him. An example why we need it: sending POOL_UPGRADE txn.

PR #57, opened 2017-03-14
test and fixes replay ledger issue

pradeep1991singh:


sovrin_node/server/node.py
```diff
@@ -565,7 +569,12 @@ def storeTxnAndSendToClient(self, reply):
          client requests it.
         """
         result = reply.result
+        if result[TXN_TYPE] in (SCHEMA, ISSUER_KEY):
+            result = deepcopy(result)
```
rajeshkalaria80: Why this line is necessary?

pradeep1991singh: @rajeshkalaria80 deepcopy can be refactored. I will use result as it is. But I need `if` condition because I need to serialize data with txn type `schema` or `issuer_key`.

PR #61, opened 2017-03-21
SOV-742: use encoded (integer) values as txn type

ashcherbakov:
- use encoded (integer) values as txn type
- rename txn.py to constants.py


sovrin_node/test/replay/test_ledger_replay.py
```diff
@@ -161,8 +160,8 @@ def testReplayLedger(addNymTxn, addedRawAttribute, submittedPublicKeys,
     nodeToStop.cleanupOnStopping = False
     nodeToStop.stop()
     looper.removeProdable(nodeToStop)
-    client = nodeToStop.graphStore.client
-    client.db_drop(client._connection.db_opened)
+    #client = nodeToStop.graphStore.client
```
dhh1128: why are these commented out?

ashcherbakov: Actually nodeToStop.stop() already drops the database, so I think the commented out lines are not needed (I had a failing test with this lines on Ubuntu).
The ticket also doesn't work on Windows. There is a ticket SOV-781 about failing of this test, I pu some comments there.

PR #77, opened 2017-03-31
returning proper error message

lovesh:


sovrin_node/test/test_suspension.py
```diff
@@ -81,9 +87,12 @@ def testTrustAnchorSuspensionByTrustee(looper, anotherTrustee, anotherTrustAncho
 
 
 def testTrusteeSuspensionByTrustee(looper, trustee, trusteeWallet,
-                                   anotherTrustee):
-    _, trWallet = anotherTrustee
+                                   anotherTrustee, anotherSteward1):
+    trClient, trWallet = anotherTrustee
     suspendRole(looper, trustee, trusteeWallet, trWallet.defaultId)
+    _, sWallet = anotherSteward1
+    with pytest.raises(AssertionError):
```
keenondrums: Shouldn't we make a separate test for this case? Something like blacklistedTrusteeCanNotBlacklistSteward

lovesh: This check is testing whether the suspension really took effect, `suspendRole` is just testing whether the suspension transaction could be made successfully

PR #79, opened 2017-03-31
Stp

ashcherbakov:
- Migration to zmq;
- secure transport protocol repo (stp) is created.


sovrin_node/server/node.py
```diff
@@ -42,13 +44,14 @@
...
 from sovrin_node.server.upgrader import Upgrader
+from stp_core.network.exceptions import EndpointException
 
 logger = getlogger()
 jsonSerz = JsonSerializer()
 
 
 class Node(PlenumNode, HasPoolManager):
-    keygenScript = "init_sovrin_raet_keep"
+    keygenScript = "init_sovrin_keys"
```
dhh1128: Have we updated any human instructions (e.g., in Getting Started) and any bash scripts that we've written that might call this command?

PR #81, opened 2017-04-03
fixed test runner

ashcherbakov:


lovesh:
@devin-fisher Its not a build tool but only a test runner, was build to get around the problem of pytest keeping all the test results, logs, etc in memory until all tests are done so if you have 300+ tests (plenum) it going to run into OOM on our build server. So this runs one module at a time so everything related to that module is gone when it ends

PR #87, opened 2017-04-06
Up common version to 23

mzk-vct:
Change required sovrin-common version in setup,py to 0.2.23


keenondrums:
Close it because of #88

PR #101, opened 2017-04-14
State Trie

lovesh:


sovrin_node/persistence/attribute_store.py
```diff
@@ -0,0 +1,25 @@
...
+
+class AttributeStore:
+    """
+    Stores attributes as key value pair where the key is hash of the
+    attribute as stored in ledger and value is the actual value if the attribute
+    """
+
+    def __init__(self, dbPath):
+        self.dbPath = dbPath
+        self.db = KVStoreLeveldb(dbPath)
```
ashcherbakov: If we need to change a KVStore type, we need to change implementation. It's better to inject `db` into constructor (in addition to `dbpah`), so that using another DB (for example RocksDB) can be done easily using existing `AttributeStore` class.

sovrin_node/persistence/idr_cache.py
```diff
@@ -0,0 +1,201 @@
...
+
+    @staticmethod
+    def decodeVerkey(verkey):
+        if verkey == IdrCache.unsetVerkey:
+            return None
+        else:
+            return verkey.decode()
+
+    @staticmethod
+    def getPrefixAndIv(guardian=None, verkey=None):
```
ashcherbakov: What is 'Iv'?

sovrin_node/persistence/idr_cache.py
```diff
@@ -0,0 +1,201 @@
...
+        iv = guardian if guardian else verkey
+        prefix = b'1' if guardian else b'0'
+        return prefix, iv
+
+    @staticmethod
+    def packIdrValue(ta=None, role=None, verkey=None):
+        # prefix, iv = IdrCache.getPrefixAndIv(guardian, verkey)
+        if ta is None:
```
ashcherbakov: What is `ta`?

sovrin_node/persistence/idr_cache.py
```diff
@@ -0,0 +1,201 @@
...
+        val = self.packIdrValue(ta, role, verkey)
+        if isCommitted:
+            self._db.set(idr, val)
+        else:
+            self.currentBatchOps.append((idr, val))
+
+    def close(self):
+        self._db.close()
+
+    def open(self):
```
ashcherbakov: If we need to change a KVStore type, we need to change implementation. It's better to inject db, so that using another DB (for example RocksDB) can be done easily without changing IdrCache class.
We can use leveldb as default for db.

sovrin_node/server/domain_req_handler.py
```diff
@@ -0,0 +1,346 @@
...
+    ISSUER_KEY, REF
+from sovrin_common.types import Request
+from stp_core.common.log import getlogger
+from stp_core.network.exceptions import EndpointException
+
+
+logger = getlogger()
+
+
+class DomainReqHandler(PHandler):
```
ashcherbakov: Avoid using the same name for classes from different modules.
Consider rename the current class to SovrinDomainReqHandler

sovrin_node/server/domain_req_handler.py
```diff
@@ -0,0 +1,346 @@
...
+    MARKER_ATTR = "\01"
+    MARKER_SCHEMA = "\02"
+    MARKER_IPK = "\03"
```
ashcherbakov: We renamed IPK to CLAIM_DEF in master recently, so please use claim_def instead

sovrin_node/server/domain_req_handler.py
```diff
@@ -0,0 +1,346 @@
...
+                if ROLE in txn:
+                    data[ROLE] = txn.get(ROLE)
+                if VERKEY in txn:
+                    data[VERKEY] = txn.get(VERKEY)
+                self.updateNym(nym, data, isCommitted=isCommitted)
+            elif typ == ATTRIB:
+                self._addAttr(txn)
+            elif typ == SCHEMA:
+                self._addSchema(txn)
+            elif typ == ISSUER_KEY:
```
ashcherbakov: We renamed IPK to CLAIM_DEF in master recently, so please use claim_def instead

sovrin_node/server/domain_req_handler.py
```diff
@@ -0,0 +1,346 @@
...
+        nym = msg.get(TARGET_NYM)
+        if self.hasNym(nym, isCommitted=False):
+            if not self.idrCache.hasTrustee(identifier, isCommitted=False) and \
+                            self.idrCache.getOwnerFor(nym, isCommitted=False) != identifier:
+                reason = '{} is neither Trustee nor owner of {}'\
+                    .format(identifier, nym)
+                return False, reason
+        return True, ''
+
+    def doStaticValidation(self, identifier, reqId, operation):
```
ashcherbakov: Please split the method to perform static validation for each txn type (like `doStaticValidationNym`, `doStaticValidationAttr`)

sovrin_node/server/domain_req_handler.py
```diff
@@ -0,0 +1,346 @@
...
+                    #                                'JSON'.format(operation[RAW]))
+
+            if not (not operation.get(TARGET_NYM) or
+                    self.hasNym(operation[TARGET_NYM], isCommitted=False)):
+                raise InvalidClientRequest(identifier, reqId,
+                                           '{} should be added before adding '
+                                           'attribute for it'.
+                                           format(TARGET_NYM))
+
+    def validate(self, req: Request, config=None):
```
ashcherbakov: Please split the method to perform validation for each txn type (like doValidateNym, doValidateAttr)

sovrin_node/server/domain_req_handler.py
```diff
@@ -0,0 +1,346 @@
...
+        if typ == NYM:
+            try:
+                originRole = s.getRole(origin, isCommitted=False) or None
+            except:
+                raise UnknownIdentifier(
+                    req.identifier,
+                    req.reqId)
+
+            nymData = s.getNym(op[TARGET_NYM], isCommitted=False)
+            if not nymData:
```
ashcherbakov: Split to `doValidateExistingNym` and `doValidateNewNym`

sovrin_node/server/domain_req_handler.py
```diff
@@ -0,0 +1,346 @@
...
+    def _makeSchemaPath(did, schemaName, schemaVersion) -> bytes:
+        return "{DID}:{MARKER}:{SCHEMA_NAME}{SCHEMA_VERSION}" \
+            .format(DID=did,
+                    MARKER=DomainReqHandler.MARKER_SCHEMA,
+                    SCHEMA_NAME=schemaName,
+                    SCHEMA_VERSION=schemaVersion) \
+            .encode()
+
+    @staticmethod
+    def _makeIssuerKeyPath(did, schemaSeqNo) -> bytes:
```
ashcherbakov: Support SIGNATURE_TYPE according to recent changes from master

sovrin_node/server/node.py
```diff
@@ -439,120 +317,62 @@ def defaultNodeAuthNr(self):
...
     def processRequest(self, request: Request, frm: str):
         if request.operation[TXN_TYPE] == GET_NYM:
-            self.processGetNymReq(request, frm)
+            self.transmitToClient(RequestAck(*request.key), frm)
+            result = self.reqHandler.handleGetNymReq(request, frm)
+            self.transmitToClient(Reply(result), frm)
+        # TODO: Come back to it
         elif request.operation[TXN_TYPE] == GET_TXNS:
-            self.processGetTxnReq(request, frm)
+            # self.processGetTxnReq(request, frm)
```
ashcherbakov: Should we come back to it and support GET_TXNS?

sovrin_node/server/node_authn.py
```diff
@@ -10,6 +10,7 @@ class NodeAuthNr(NaclAuthNr):
     def __init__(self, ledger: Ledger):
         self.ledger = ledger
 
+    # TODO: This is not right, what if verkey changes?
```
ashcherbakov: It's very inefficient to get all transactions and iterate through them. Do we use this method? If we use it, then it needs to be fixed.

sovrin_node/server/pool_req_handler.py
```diff
@@ -0,0 +1,64 @@
...
+
+from plenum.common.ledger import Ledger
+from plenum.persistence.pruning_state import PruningState
+from plenum.server.pool_req_handler import PoolRequestHandler as PHandler
+from sovrin_common.auth import Authoriser
+from sovrin_common.constants import NODE
+from sovrin_node.persistence.idr_cache import IdrCache
+
+
+class PoolRequestHandler(PHandler):
```
ashcherbakov: Please use a different class name (avoid classes with the same name from different packages)

sovrin_node/test/helper.py
```diff
@@ -241,16 +241,10 @@ def getUpgrader(self):
...
-        if not hasattr(self, '_orientDbStore'):
-            self._orientDbStore = orientdb_store.createOrientDbInMemStore(
-                self.config, name, dbType)
-        return self._orientDbStore
-
     def onStopping(self, *args, **kwargs):
         if self.cleanupOnStopping:
             self.cleanupDataLocation()
-        self.graphStore.store.close()
+        # self.graphStore.store.close()
```
ashcherbakov: Do we clean KV storages on stopping?

sovrin_node/test/replay/test_ledger_replay.py
```diff
@@ -1,179 +1,179 @@
...
-    looper.add(newNode)
-    nodeSet[0] = newNode
-    looper.run(checkNodesConnected(nodeSet, customTimeout=30))
-    looper.run(eventually(checkNodeLedgersForEquality, newNode,
-                          *txnPoolNodeSet[1:4], retryWait=1, timeout=15))
-
-    compareGraph("NYM", nodeSet)
-    compareGraph("IssuerKey", nodeSet)
-    compareGraph("Schema", nodeSet)
+# import pytest
```
ashcherbakov: Please remove the test if it isn't needed anymore

sovrin_node/persistence/idr_cache.py
lovesh: Obsolete, remvoed

sovrin_node/persistence/idr_cache.py
lovesh: Trust Anchor

sovrin_node/server/node.py
lovesh: No, its fine, i was premature to build it

sovrin_node/test/helper.py
lovesh: Only for tests, its done when you clean directory of the node

sovrin_node/server/node_authn.py
lovesh: It will just go over the pool ledger and this happens only in case of succssfull code upgrade but yes, it can be improved

sovrin_node/server/domain_req_handler.py
lovesh: We should avoid leaking details of base classes, its a request handler and in most cases we should not bother the devs with details like its base class, also people dont need to use Plenum's pool request handler unless they are extending it

sovrin_node/server/pool_req_handler.py
lovesh: We should avoid leaking details of base classes, its a request handler and in most cases we should not bother the devs with details like its base class, also people dont need to use Plenum's pool request handler unless they are extending it

sovrin_node/persistence/idr_cache.py
ashcherbakov: Please use the full name for variables. it's not a common abbreviation, and not clear for the reader of the code

sovrin_node/server/domain_req_handler.py
ashcherbakov: What details do we expose by renaming? It's easy to make a mistake and import a wrong class (from Plenum) if they have the same name.

sovrin_node/server/node.py
ashcherbakov: Does it mean that GET_TXNS doesn't work now?

sovrin_node/server/pool_req_handler.py
ashcherbakov: What details do we expose by renaming? It's easy to make a mistake and import a wrong class (from Plenum) if they have the same name.

sovrin_node/server/domain_req_handler.py
lovesh: You are telling a dev that this is a SovrinRequestHandler and hinting that there is another handler, a dev cannot use that handler anywhere unless he gets rid of this and creates a new one from the base. Why do all that?

sovrin_node/server/node.py
lovesh: It did not work before either, its been inactive for quiet some time

sovrin_node/server/domain_req_handler.py
ashcherbakov: Plenum and Sovrin are separate products. If RequestHandler is part of Plenum's API, then dev must know that there are different implementations.
I don't think it's good to have parent and child class have the same name

PR #131, opened 2017-04-25
Upgrader fix

spivachuk:
- Fixed a bug in Upgrader.compareVersions method with truncating trailing zeroes in the version number. For example, the version number "0.3.70" was previously truncated to "0.3.7".
- Enhanced testVersions test to verify the fix.


sovrin_node/test/upgrade/test_upgrader.py
```diff
@@ -7,6 +7,8 @@ def testVersions():
     assert not Upgrader.isVersionHigher('0.0.9', '0.0.8')
```
lovesh: @spivachuk Can you please replace this `isVersionHigher` with semver comparisons like `semver.match("0.3.70", ">0.3.200")` etc, we are not going to be moving away from semver so lets use the library and get rid of our code.

PR #159, opened 2017-05-30
Feature/migrations

keenondrums:


data/migrations/README.md
```diff
@@ -0,0 +1,16 @@
+# Migrations for sovrin-node
+
+All migrations are python scripts with names satisfying the following rule:
+```
+<underscored_version-migration-should-be-applied-to>_<migration-name>
```
dhh1128: I am feeling concerned about two issues with this approach:
1. This creates multiple migration scripts per version change. I want one migration script per update. Anything more is unnecessary complexity. Also, it gives multiple points of failure instead of one, and it gives no responsibility to these individual scripts to know about or coordinate with one another. I think it increases the chances of failure.
2. This mechanism doesn't include in its name the version you are coming from, only the version you are headed to. I think this is problematic because we might have nodes upgrading from more than one starting point.

dhh1128: Why are versions underscored here, but use dots in .deb? Can we use dots here?

scripts/migration_tool.py
```diff
@@ -0,0 +1,52 @@
...
+
+logger = getlogger()
+
+
+# Returns number of performed migrations
+def migrate(current_version):
+    logger.info('Migrating from {}'.format(current_version))
+    migration_scripts = _get_migration_scripts()
+    logger.debug('Found migration scripts: {}'.format(migration_scripts))
+    migration_scripts = _get_relevalnt_migrations(migration_scripts, current_version)
```
dhh1128: type "relevalnt"

scripts/migration_tool.py
```diff
@@ -0,0 +1,52 @@
...
+        migration_split = migration.split('_')
+        migration_version = '.'.join(migration_split[0:3])
+        if migration_version >= current_version:
+            relevant_migrations.append(migration)
+    relevant_migrations.sort()
+    return relevant_migrations
+
+
+def _call_migration_script(migration_script):
+    importlib.import_module(SCRIPT_PREFIX + migration_script)
```
dhh1128: How do we account for differences in import path across upgrades?

scripts/node_control_tool.py
```diff
@@ -14,30 +19,41 @@ def compose_cmd(cmd):
...
-            retcode = subprocess.call(compose_cmd(['upgrade_sovrin_node_test', version]), shell=True)
-        else:
-            retcode = subprocess.call(compose_cmd(['upgrade_sovrin_node', version]), shell=True)
-        if retcode != 0:
-            print('Upgrade failed')
-    except:
-        print('Something went wrong with calling upgrade script')
+    logger.info('Upgrading sovrin node to version %s, test_mode %d ' % (version, int(test_mode)))
+    if test_mode:
+        retcode = subprocess.call(compose_cmd(['upgrade_sovrin_node_test', version]), shell=True)
```
dhh1128: subprocess.call(shell=True) is a security risk.

dhh1128: What privileges are in effect when this migration mechanism runs? Does it delegate those privileges to child processes? Should it?

data/migrations/README.md
keenondrums: @dhh1128 
1) I think we will have a single script almost always. yet I think we need to leave some room for us to wiggle, since it does not cost us anything.
2) It's actually the version you are coming from. I think it does not really matter what version you're updating to
3) Versions are underscored because they are python modules and a python module can not be dotted

scripts/migration_tool.py
keenondrums: @dhh1128 I guess in the future we will switch SCRIPT_PREFIX based on operating system

scripts/node_control_tool.py
keenondrums: @dhh1128 node_control_tool.py is run under root. We need root privileges for upgrade because we install new packages with apt-get.

dhh1128: node_control_tool.py should NOT be run as root (uid==0); it should be run as an account that has elevated privileges to issue calls to apt-get and restart daemons.

scripts/migration_tool.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+def _get_relevalnt_migrations(migration_scripts, current_version):
+    relevant_migrations = []
+    for migration in migration_scripts:
+        migration_split = migration.split('_')
+        migration_version = '.'.join(migration_split[0:3])
+        if migration_version >= current_version:
+            relevant_migrations.append(migration)
+    relevant_migrations.sort()
```
keenondrums: I think each migration should be cross-platform. As to case-sensivity it does not really matter because we will rarely have several migrations for one version. Even if we have then we have a section in readme with recomendations to prefix migration names with digits

keenondrums: I think each migration should be cross-platform. As to case-sensivity it does not really matter because we will rarely have several migrations for one version. Even if we have then we have a section in readme with recomendations to prefix migration names with digits

data/migrations/README.md
dhh1128: I don't agree with the idea that "it does not cost us anything." Please collapse to 1 script and eliminate the sorting/collection. The cost of more than 1 script is complexity, as I pointed out. It's not just code complexity--it's communication/doc complexity, and complexity in testing/managing the coordination actions across multiple scripts.

Re #2: versioning needs to include a src and a target version. Of these two, the *target* is more important. We can't say, "give me a script that upgrades version X to any version later than that" -- it's just not practical. Instead, we want to say, "give me a script that upgrades X to Y". Then we can play all the migrations in order to get from X to Z.

scripts/node_control_tool.py
keenondrums: Apt-get requires root. https://askubuntu.com/questions/339/how-can-i-install-a-package-without-root-access
Even if we can find a way around it's a different issue then, isn't it?

scripts/node_control_tool.py
```diff
@@ -14,30 +19,41 @@ def compose_cmd(cmd):
...
+    else:
+        retcode = subprocess.call(compose_cmd(['upgrade_sovrin_node', version]), shell=True)
+    if retcode != 0:
+        msg = 'Upgrade failed: upgrade script returned {}'.format(retcode)
```
keenondrums: Right now a steward should receive an email prior to the upgrade if he has email notifier installed. Then he should receive an email with the result of the upgrade once the node started. If there's a failure he will contact dev department with the logs

keenondrums: Upgrade script is a script which performs the installation of new code. It's always the same. Migration scripts are optional, they can be applied after upgrade if some major changes appeared in the code, backwards compatibility was broken and system needs to update its state to work with the new code

scripts/node_control_tool.py
dhh1128: Apt-get does not require root (uid==0). It requires elevated privileges (possibly euid==0). In the past 5 years, I have probably run apt-get 1000 times--5 as root, and 995 via sudo. I believe that sometimes I have run with sudo via a user who had limited privs, not ALL in the sudoers file, which would mean it is possible to create a user that just has the ability to run "sudo apt-get ..." in the sudoers file, nothing else.

You are correct that this is a different issue--but I am irritated that we are being sloppy about privileges. Please log a separate ticket.

scripts/node_control_tool.py
dhh1128: Thanks for the clarification.

PR #175, opened 2017-06-08
change idr cache to store batches in list to avoid problems with batches that result to same state roots

lovesh:


sovrin_node/persistence/idr_cache.py
```diff
@@ -91,18 +94,28 @@ def close(self):
...
+        # Commit an already created batch
+        if self.unCommitted:
+            assert self.unCommitted[0][0] == stateRoot, 'The first created batch has ' \
+                                                     'not been committed or ' \
+                                                     'reverted and yet another ' \
+                                                     'batch is trying to be ' \
+                                                     'committed'
+            self._keyValueStorage.setBatch([(idr, val) for idr, val in
+                                            self.unCommitted[0][1].items()])
+            self.unCommitted = self.unCommitted[:-1]
```
ashcherbakov: Shouldn't it be `self.unCommitted = self.unCommitted[1:]` so that we remove the first (processed item), not the last one.

sovrin_node/persistence/idr_cache.py
```diff
@@ -91,18 +94,28 @@ def close(self):
         self._keyValueStorage.close()
 
     def currentBatchCreated(self, stateRoot):
```
ashcherbakov: Can we write a more low-level test that ensures that a sequence of 
currentBatchCreated, batchRejected and onBatchCommitted calls work?
Example:
Test1: - sunny day
- save current state and current unCommitted 
- call currentBatchCreated twice (with different states)
- call onBatchCommitted for the first state
- make sure the state is committed and the unCommitted state is equal to the second state
- call onBatchCommitted for the second state
- make sure the state is committed and unCommitted is empty

Test2: - commit not the first state
- save current state and current unCommitted 
- call currentBatchCreated twice (with different states)
- call onBatchCommitted for the second state
- make sure assertion error is raised

Test3: - reject state
- save current state and current unCommitted 
- call currentBatchCreated twice (with different states)
- call batchRejected 
- make sure the second state is rejected

test4:
-

sovrin_node/persistence/idr_cache.py
```diff
@@ -91,18 +94,28 @@ def close(self):
...
-        if stateRoot:
-            self.unCommitted[stateRoot] = OrderedDict(self.currentBatchOps)
-        else:
-            self.currentBatchOps = []
+    def batchRejected(self):
+        # Batches are always rejected from end of `self.unCommitted`
+        self.currentBatchOps = []
+        self.unCommitted = self.unCommitted[:-1]
 
     def onBatchCommitted(self, stateRoot):
```
ashcherbakov: Why do we assume in the API that batches are always rejected from the end of self.unCommitted and do not pass state into `batchRejected`, but pass the state explicitly into `onBatchCommitted` and check for assertion?
Shouldn't we either pass state into both methods and check assertions in both method (that the state is equal to the first batch's state in onBatchCommitted and is equal to the last batch's state in batchRejected), or don't pass it at all?

sovrin_node/persistence/idr_cache.py
lovesh: Yes, fixed

sovrin_node/persistence/idr_cache.py
lovesh: Yes, Test1 can be done. To Test2 there has to be scenario created which will result in calling `revert` in an incorrect order of batches, simply patching some method to deliberately call `onBatchCommitted` with wrong args only proves that an assertion check works, its ok to do it, but not worth investing time in. I would rather create a scenario that result in reverting batches in incorrect order and see failures much before this code is reached. Test3 yes can be done

sovrin_node/persistence/idr_cache.py
lovesh: We assume that revert in reverse order is guaranteed by the replica
Because we do not preserve state-root after a pre-prepare in replica, so if we preserve it somewhere then we can pass that to replica and before reverting check it.

PR #194, opened 2017-06-26
INDY-260: Added parallel read-write tests

spivachuk:
- Added a test of cyclic updating own verkeys by some users and cyclic reading them by other users in parallel.
- Added a test of cyclic updating and reading own verkeys by multiple users in parallel.


spivachuk:
Closed this PR because moved the parallel read-write system tests from sovrin-node project to sovrin-client project.

PR #197, opened 2017-06-28
INDY-156 POOL_UPGRADE txn unclear error message

hadleym:
Added logic to sovrin_node/server/config_req_handler.py
When a POOL_UPGRADE request that has already been accepted is submitted, there will be a UnauthorizedClientRequest thrown now.


sovrin_node/server/config_req_handler.py
```diff
@@ -63,6 +63,12 @@ def validate(self, req: Request, config=None):
             # present
             status = self.upgrader.statusInLedger(req.operation.get(NAME),
                                                   req.operation.get(VERSION))
+            if status == START and action == START:
+                raise UnauthorizedClientRequest(
```
ashcherbakov: Why do we raise `UnauthorizedClientRequest` here? Should we use some other type of exception? `UnauthorizedClientRequest` is usually raised when a sender doesn't have enough permissions (for example Role) to perform a request.

hadleym: Sure, we can make it a more appropriate message.  The ticket suggested UnauthorizedClientRequest, I will change and re-submit this morning.

hadleym:
changed to InvalidClientRequest

sovrin_node/test/upgrade/test_pool_upgrade_reject.py
```diff
@@ -42,4 +42,10 @@ def testNonTrustyCannotCancelUpgrade(looper, validUpgradeSent,
     looper.run(eventually(checkRejects, stClient, req.reqId,
                           'cannot do'))
 
+def test_accept_then_reject_upgrade(looper, trustee, trusteeWallet, validUpgradeSent, validUpgrade):
+    validUpgrade2 = deepcopy(validUpgrade)
+    _, req = sendUpgrade(trustee, trusteeWallet, validUpgrade2)
+    timeout = plenumWaits.expectedReqNAckQuorumTime()
+    looper.run(eventually(checkRejects, trustee, req.reqId,
+                          'UnauthorizedClientRequest', retryWait=1, timeout=timeout))
```
ashcherbakov: Should we change this to be `InvalidClientRequest`?

hadleym: This was changed to InvalidClientRequest.

hadleym:
Reposted PR in slack. Please let me know if this is not the correct work flow.

sovrin_node/test/upgrade/test_pool_upgrade_reject.py
hadleym: Ah, I was not correcting this properly.

PR #205, opened 2017-07-03
Support GET_TXN transaction

Artemkaaas:


sovrin_node/server/domain_req_handler.py
```diff
@@ -235,6 +235,21 @@ def handleGetSchemaReq(self, request: Request, frm: str):
         }}
         return result
 
+    def handleGetTnxReq(self, request: Request, frm: str, node):
```
lovesh: This method needs to be defined on Node and not a `ReqHandler` since a `ReqHandler` corresponds to one ledger only but any ledger can be queried. The client request needs to indicate which ledger needs to be queried for the sequence number. Also this method should be in plenum.

mzk-vct:
Not relevant anymore, following will be merged instead https://github.com/hyperledger/indy-node/pull/210

PR #228, opened 2017-07-13
Increased testing criteria

hadleym:
Made the testing message a bit better by adding the upgrade NAME to the test, verifying that the appropriate exception was actually being caught.


hadleym:
INDY-156

PR #250, opened 2017-07-20
Add new command POOL_CONFIG

dsurnin:


sovrin_node/test/pool_config/test_send_pool_config_only_trustee.py
```diff
@@ -0,0 +1,29 @@
+from sovrin_client.test.helper import getClientAddedWithRole
+from sovrin_client.test.helper import checkRejects
+from stp_core.loop.eventually import eventually
+from sovrin_node.test.pool_config.helper import ensurePoolConfigSent, checkPoolConfigWritableSet, sendPoolConfig
+from plenum.common.constants import STEWARD
+
+
+def test_only_trustee_send_pool_config_wtff(nodeSet, tdir, looper, trustee, trusteeWallet, poolConfigWTFF):
```
ashcherbakov: Please use full names (write_force, not_write_not_force, etc.)

sovrin_node/server/node.py
```diff
@@ -365,12 +379,19 @@ def processRequest(self, request: Request, frm: str):
...
         else:
             # forced request should be processed before consensus
-            if request.operation[TXN_TYPE] == POOL_UPGRADE and request.isForced():
+            if (request.operation[TXN_TYPE] in [POOL_UPGRADE, POOL_CONFIG]) and request.isForced():
                 self.configReqHandler.validate(request)
                 self.configReqHandler.applyForced(request)
-            super().processRequest(request, frm)
+            #here we should have write transactions that should be processed only on writable pool
+            if self.poolCfg.isWritable() or (request.operation[TXN_TYPE] == POOL_CONFIG):
+                super().processRequest(request, frm)
```
ashcherbakov: Do we have tests to check that read (GET) request can be processed regardless of writeable state?

dsurnin: Get Nym test aded

sovrin_node/server/node.py
```diff
@@ -86,13 +87,28 @@ def __init__(self,
...
         self.initConfigState()
         self.requestExecuter[CONFIG_LEDGER_ID] = self.executeConfigTxns
 
         self.nodeMsgRouter.routes[Request] = self.processNodeRequest
         self.nodeAuthNr = self.defaultNodeAuthNr()
 
+        self.nodeMsgRouter.routes[PrePrepare] = self.extraPrePrepareProcesing
+
+
+    def extraPrePrepareProcesing(self, msg, frm):
```
ashcherbakov: Does it replace default PrePrepare handler?

ashcherbakov: Please add a test to check that PrePrepares are not processed during read-only state unless this is a CONFIG_LEDGER.
Do we process POOL_UPGRADE when read-only? Please add a test

PR #270, opened 2017-07-28
INDY-410 - stewards can demote their own nodes

glowkey:
use command:
"send NODE dest=<NODE ID> data={'alias': '<node name>', 'services': []}"


sovrin_client/test/cli/test_node_demotion.py
```diff
@@ -0,0 +1,33 @@
...
+    be(newStewardCli)
+    do('send NODE dest={newNodeIdr} data={newNodeData}',
+       mapper=newNodeVals, expect=NODE_REQUEST_COMPLETED, within=8)
+
+    newNodeVals['newNodeData'][SERVICES] = []
+
+    do('send NODE dest={newNodeIdr} data={newNodeData}',
+       mapper=newNodeVals, expect=NODE_REQUEST_COMPLETED, within=8)
+
+    ensurePoolIsOperable(be, do, newStewardCli)
```
ashcherbakov: How do we check that the Node is really demoted?

PR #281, opened 2017-08-04
INDY-378: Corrected getting of values from repo

spivachuk:
- Corrected getting of values from SovrinPublicRepo. SchemaNotFoundError and ValueError are now raised in case the value is not found (for schemas and other values correspondingly).
- Corrected existing tests correspondingly to the made changes.
- Added tests for send CLAIM_DEF command which verify the fix.
- Updated indy-anoncreds dependency.


sovrin_client/test/anon_creds/test_public_repo.py
```diff
@@ -112,14 +114,16 @@ def testGetSchemaBySeqNo(submittedSchemaDefGvt, publicRepo, looper):
 
 
 def testGetSchemaByInvalidSeqNo(submittedSchemaDefGvt, publicRepo, looper):
```
ashcherbakov: Do we have a test in CLI reproducing the issue from INDY-378? Does the CLI show exception properly in this case not crashing?

spivachuk: Yes, with this fix the CLI shows errors with a wrong `ref` argument properly. Wrote the tests verifying the CLI output in such cases.

PR #284, opened 2017-08-07
Indy 316

keenondrums:


sovrin_node/server/upgrader.py
```diff
@@ -76,6 +80,17 @@ def versionsDescOrder(versions):
         return sorted(versions,
                       key=cmp_to_key(Upgrader.compareVersions))
 
+    @staticmethod
+    def get_upgrade_id(txn):
+        seq_no = txn.get(F.seqNo.name, '')
+        if txn.get(FORCE, None):
```
ashcherbakov: Do we have tests that there is no loop with FORCE=True?

PR #287, opened 2017-08-08
Walleted.py merge_request Did Not Work

hadleym:
Positional arguments for the else case when proof
requests already existed case in
_merge_request were not correct and caused
the attributes member to not be created
correctly.

### DIFF
--- a/sovrin_client/test/cli/test_merge_invitation.py
+++ b/sovrin_client/test/cli/test_merge_invitation.py
@@ -1,0 +1,29 @@
+import pytest
+
+from sovrin_client.agent.agent_issuer import AgentIssuer
+from sovrin_client.agent.agent_prover import AgentProver
+from sovrin_client.agent.agent_verifier import AgentVerifier
+from sovrin_client.agent.walleted import Walleted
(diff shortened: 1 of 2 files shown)


sovrin_client/test/cli/test_merge_invitation.py
```diff
@@ -0,0 +1,29 @@
...
+from sovrin_client.agent.agent_prover import AgentProver
+from sovrin_client.agent.agent_verifier import AgentVerifier
+from sovrin_client.agent.walleted import Walleted
+from sovrin_client.client.wallet.connection import Connection
+from sovrin_client.client.wallet.wallet import Wallet
+from sovrin_client.agent.walleted_agent import WalletedAgent
+from plenum.common.constants import NAME, VERSION, ATTRIBUTES, VERIFIABLE_ATTRIBUTES
+
+
+def test_merge_invitation():
```
ashcherbakov: The test checks merging to an empty link only. Can we also check that merging works for non-empty link (real merging?)

hadleym:
Updated to have more testing of the merge_invitation function.

PR #297, opened 2017-08-10
fixes for batching "send NYM" requests (INDY-292)

glowkey:
Adds second test to verify batches work when NYMs are different and batch fails when NYMs are the same

### DIFF
--- a/sovrin_client/test/cli/test_nym.py
+++ b/sovrin_client/test/cli/test_nym.py
@@ -255,7 +255,7 @@ def testNewKeyChangesWalletsDefaultId(be, do, poolNodesStarted, poolTxnData,
        expect=["Nym {} added".format(idr)])
 
 
-def testSend2NymsSucceedsWhenBatched(
+def test_send_same_nyms_fails_when_batched(
         be, do, poolNodesStarted, newStewardCli):
 
     be(newStewardCli)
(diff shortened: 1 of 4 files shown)


sovrin_node/persistence/idr_cache.py
```diff
@@ -73,10 +73,12 @@ def get(self, idr, isCommitted=True):
         else:
             # Looking for uncommitted values, iterating over `currentBatchOps and unCommitted`
             # in reverse to get the latest value
-            for _, cache in reversed(self.currentBatchOps):
-                if idr in cache:
-                    value = cache[idr]
-                    break;
+            for iter, cache in reversed(self.currentBatchOps):
```
ashcherbakov: `iter` is a key word. Should it be `idr`?

sovrin_client/test/cli/test_nym.py
```diff
@@ -282,3 +282,42 @@ def testSend2NymsSucceedsWhenBatched(
     do('send GET_NYM dest={dest}',
         mapper=parameters, expect=CURRENT_VERKEY_FOR_NYM, within=2)
 
+def test_send_different_nyms_succeeds_when_batched(
```
ashcherbakov: How is it related to batching? Do we send commands in the same batch?

sovrin_client/test/cli/test_nym.py
```diff
@@ -255,7 +255,7 @@ def testNewKeyChangesWalletsDefaultId(be, do, poolNodesStarted, poolTxnData,
        expect=["Nym {} added".format(idr)])
 
 
-def testSend2NymsSucceedsWhenBatched(
+def test_send_same_nyms_fails_when_batched(
```
ashcherbakov: Why the tests is called `fails`? It looks like the test checks a positive case.

sovrin_node/persistence/idr_cache.py
glowkey: I've changed that.

sovrin_client/test/cli/test_nym.py
glowkey: there are 2 requests, one with "entercmd" and the second with "do".  "entercmd" doesn't send it to the server so when "do" executes it will have both in a batch.  I've entered comments to clarify this.

sovrin_client/test/cli/test_nym.py
glowkey: The second request in the batch shouldn't be written (the same behavior as when not batched).  The test verifies this by making sure the first request is what is written.  I've renamed the test to clarify this.

PR #299, opened 2017-08-11
Closes INDY-251, INDY-259

keenondrums:
Bug Fixed: POOL_UPGRADE only installs the latest build. You are not able to specify a version other than the latest to upgrade to.

Feature Added: POOL_UPGRADE should support downgrades

### DIFF
--- a/sovrin_node/test/upgrade/test_upgrader.py
+++ b/sovrin_node/test/upgrade/test_upgrader.py
@@ -3,12 +3,10 @@ from sovrin_node.server.upgrader import Upgrader
 
 
 def testVersions():
-    assert Upgrader.isVersionHigher('0.0.5', '0.0.6')
-    assert not Upgrader.isVersionHigher('0.0.9', '0.0.8')
-    assert Upgrader.isVersionHigher('0.0.9', '0.1.0')
-    assert Upgrader.isVersionHigher('0.20.30', '1.0.0')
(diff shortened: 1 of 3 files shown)


sovrin_node/test/upgrade/test_upgrader.py
```diff
@@ -3,12 +3,10 @@
...
-    assert Upgrader.isVersionHigher('0.0.5', '0.0.6')
-    assert not Upgrader.isVersionHigher('0.0.9', '0.0.8')
-    assert Upgrader.isVersionHigher('0.0.9', '0.1.0')
-    assert Upgrader.isVersionHigher('0.20.30', '1.0.0')
-    assert Upgrader.isVersionHigher('1.3.19', '1.3.20')
-    versions = ['0.0.1', '0.10.11', '0.0.10', '0.0.2',
-                '1.9.0', '9.10.0', '9.1.0']
-    assert Upgrader.versionsDescOrder(versions) == \
-           ['9.10.0', '9.1.0', '1.9.0', '0.10.11', '0.0.10', '0.0.2', '0.0.1']
+    assert Upgrader.compareVersions('0.0.5', '0.0.6') == 1
```
ashcherbakov: Why do we check patch versions only? Please have tests for comparison of different major and minor versions as well.

PR #338, opened 2017-09-06
Hotfix/migration to 1.1.33 (INDY-808) (#337)

andkononykhin:
cherry-picked 43c83dc

* added test for _get_migration_scripts function

* fixed path resolving for python modules

### DIFF
--- a/sovrin_node/test/upgrade/test_migration_tool.py
+++ b/sovrin_node/test/upgrade/test_migration_tool.py
@@ -45,3 +45,12 @@ def testMigrateTimesOut(monkeypatch):
 
     with pytest.raises(TimeoutError):
         migration_tool.migrate(TEST_VERSION, TEST_NEW_VERSION, TEST_TIMEOUT)
+
+def testGetMigrationScripts():
+    try:
+        res = migration_tool._get_migration_scripts(migration_tool._get_current_platform())
+    except Exception as e:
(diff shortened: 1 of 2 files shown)


sovrin_node/test/upgrade/test_migration_tool.py
```diff
@@ -45,3 +45,12 @@ def testMigrateTimesOut(monkeypatch):
...
     with pytest.raises(TimeoutError):
         migration_tool.migrate(TEST_VERSION, TEST_NEW_VERSION, TEST_TIMEOUT)
+
+def testGetMigrationScripts():
+    try:
+        res = migration_tool._get_migration_scripts(migration_tool._get_current_platform())
+    except Exception as e:
+        pytest.fail("Unexpected error: {}".format(e))
+    else:
+        assert type(res) is list
```
lovesh: Is there a guarantee that `res` is always a `list` and not a subclass of `list`? If not then replace by `assert isinstance(res, list)`

andkononykhin: yes, currently list contructor is used directly https://github.com/andkononykhin/indy-node/blob/5fa2f762d479e68a8e02e2b59b94b4c2829284d8/sovrin_node/utils/migration_tool.py#L76
I think it could be adjusted in future if something changed and test fails.

PR #340, opened 2017-09-07
fix pointers to old (bad) instructions

trentlarson:
Note that there are also references to sovrin-foundation/sovrin-common which maybe should now be hyperledger/indy-common.


trentlarson:
Note that there are also references in these files to sovrin-foundation/sovrin-common, which I'm guessing should be hyperledger/indy-common.

PR #343, opened 2017-09-12
Feature/optional looper debug

sergey-shilov:


scripts/load_test.py
```diff
@@ -392,7 +392,7 @@ def buildCoros(coroFunc, corosArgs):
     clientPoll = ClientPoll(args.clientsListFilePath,
                             args.numberOfClients, args.numberOfClientsToSkip)
 
-    with Looper(debug=True) as looper:
+    with Looper() as looper:
```
andkononykhin: Why config.LOOPER_DEBUG is not used here?

sergey-shilov: Because we decided to use default debug value of Looper class (False) or use False explicitly for testing functionality. As I understand, 'scripts/load_test.py' script is related to tests.

PR #359, opened 2017-09-26
INDY-833: add migration script

sergey-shilov:


data/migrations/deb/x_x_xx_to_y_y_yy.py
```diff
@@ -0,0 +1,79 @@
...
+    old_daemon_config = os.path.join(old_daemon_config_base_dir, 'indy.env')
+
+    new_log_dir = os.path.join(config.LOG_DIR, config.NETWORK_NAME)
+    new_node_base_dir = os.path.join(base_dir , config.NETWORK_NAME)
+    new_node_base_data_dir = os.path.join(config.NODE_BASE_DATA_DIR, config.NETWORK_NAME)
+
+    new_general_config = os.path.join(config.GENERAL_CONFIG_DIR, config.GENERAL_CONFIG_FILE)
+
+    # Move general config file
+    os.rename(old_general_config, new_general_config)
```
keenondrums: why rename?

data/migrations/deb/x_x_xx_to_y_y_yy.py
```diff
@@ -0,0 +1,79 @@
...
+    shutil.move(old_nodes_data_dir, new_node_base_data_dir)
+
+    # Move wallets directory
+    shutil.move(old_wallets_dir, new_node_base_dir)
+
+    # Move daemon config
+    shutil.move(old_daemon_config, config.GENERAL_CONFIG_DIR)
+
+
+if os.path.exists(old_base_dir):
```
keenondrums: if we do not find the directory I think we should raise an exception and therefore roll back

data/migrations/deb/x_x_xx_to_y_y_yy.py
```diff
@@ -0,0 +1,79 @@
...
+
+def migrate_genesis_txn(old_base_txn_dir, new_base_txn_dir):
+    for suffix in ('sandbox', 'live', 'local'):
+        old_domain_genesis = os.path.join(
+            old_base_txn_dir, 'domain_transactions_{}_genesis'.format(suffix))
+        old_pool_genesis = os.path.join(
+            old_base_txn_dir, 'pool_transactions_{}_genesis'.format(suffix))
+
+        if os.path.exists(old_domain_genesis):
+            shutil.move(old_domain_genesis, new_base_txn_dir)
```
keenondrums: It once mentioned that they do not wanna do anything disruptive like moving, they wanna do copying instead. Please write an email to clarify

sergey-shilov: I checked this point with Alexander earlier and we decided to use 'move' strategy instead of 'copy'. Who is the person for clarification? Can I leave as is for now?

data/migrations/deb/x_x_xx_to_y_y_yy.py
sergey-shilov: I can replace it with shutil.move(), but in fact shutil.move() calls os.rename() if the destination is on the current filesystem (our case, I think). But anyway, ok, I'll use shutil.move() here.

data/migrations/deb/x_x_xx_to_y_y_yy.py
sergey-shilov: Done.

PR #369, opened 2017-10-02
INDY-878: re-factor configuration files

sergey-shilov:
- Added platform configs and base user config file.
 - Changed build prepare procedure to change/merge configs.

### DIFF
--- a/build-scripts/ubuntu-1604/prepare-package.sh
+++ b/build-scripts/ubuntu-1604/prepare-package.sh
@@ -10,6 +10,8 @@ version_dotted="$2"
 METADATA_FNAME="__metadata__.py"
 MANIFEST_FNAME="manifest.txt"
 
+USER_CONFIG_DIR="\/etc\/indy"
+
 echo -e "\n\nAbout to start updating package $repo to version $version_dotted info from cur dir: $(pwd)"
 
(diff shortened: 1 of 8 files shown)


build-scripts/ubuntu-1604/prepare-package.sh
```diff
@@ -48,4 +50,14 @@ echo "Adding manifest\n=======\n$manifest\n=======\n into $manifest_file"
 rm -rf $manifest_file
 echo -e $manifest >$manifest_file
 
+echo "Preparing config files"
+# Define user config directory
+sed -i "s/^\(USER_CONFIG_DIR\s*=\s*\).*\$/\1\"$USER_CONFIG_DIR\"/" "$repo/sovrin_common/config.py"
+# Create user config
+cp $repo/sovrin_node/user_config/user_config.py $repo/sovrin_node/user_config/indy_config.py
```
ashcherbakov: Should we copy it to USER_CONFIG_DIR=/etc/indy?

sergey-shilov: No, it is the script that prepares sources for packaging, this 'cp' is the first instruction for merging of user config and platform config into single config file indy_config.py: user config is the base (cp), platform config is appended to base (cat >>).

PR #374, opened 2017-10-05
Improve packaging steps in preparation to CentOS 7

btlogy:
- Quote shell variables to support space in paths (always good).
- Work around missing "docker volume rm --force" option in older API (< 1.25). So it also work with docker 1.12 provided with CentOS 7 (should not hurt).
- Support parallel docker build by adding suffix to docker volumes in case of multi-branch pipeline (should not change anything for single-branch).

REM: I'll PR similar improvements in sovrin, plenum and anoncreds repos if this one is merged.


btlogy:
Big mess while trying to add the Sign-Off...

PR #379, opened 2017-10-06
INDY-877: protocol version support in indy-node

ashcherbakov:

### DIFF
--- a/indy_node/server/domain_req_handler.py
+++ b/indy_node/server/domain_req_handler.py
@@ -425,7 +425,7 @@ class DomainReqHandler(PHandler):
             TXN_TIME: update_time
         }}
         if request.protocolVersion and \
-                        request.protocolVersion >= PlenumProtocolVersion.STATE_PROOF_SUPPORT.value:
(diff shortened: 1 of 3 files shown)


indy_node/server/domain_req_handler.py
```diff
@@ -421,8 +422,11 @@ def make_result(request, data, last_seq_no, update_time, proof):
             f.IDENTIFIER.nm: request.identifier,
             f.REQ_ID.nm: request.reqId,
             f.SEQ_NO.nm: last_seq_no,
-            TXN_TIME: update_time,
-            STATE_PROOF: proof
+            TXN_TIME: update_time
         }}
+        if request.protocolVersion and \
```
mzk-vct: As of now it is okay, but more strict approach should be considered in a future or we risk to get complex code with lots of if-elses here and there one day.

ashcherbakov: I agree. The problem is that each protocol version will affect different parts of code (this versions deals with state proofs; the next version may deal with serialization for example).
It would be great to have one place with all the rules (I have feeling that it could not be done easily), or to have, for example, some annotations marking the minimal version the code (method, class) can be run for.

PR #390, opened 2017-10-09
INDY-231: Fixed a bug with wrong upgrade time

spivachuk:
- Fixed a bug with a wrong time the upgrade was scheduled on.
- Added a test for the fix.


indy_client/test/cli/test_pool_upgrade_schedule.py
```diff
@@ -0,0 +1,26 @@
...
+import pytest
+from datetime import timedelta
+
+from indy_client.test.cli.test_pool_upgrade import poolUpgradeSubmitted
+from indy_client.test.cli.test_pool_upgrade import poolUpgradeScheduled
+from indy_node.test.upgrade.conftest import validUpgrade as _validUpgrade
+
+
+@pytest.fixture(scope='module')
+def validUpgrade(_validUpgrade):
```
ashcherbakov: Please use snake case

indy_client/test/cli/test_pool_upgrade_schedule.py
```diff
@@ -0,0 +1,26 @@
...
+    # of send POOL_UPGRADE command
+    upgradeSchedule = _validUpgrade['schedule']
+    for nodeId in upgradeSchedule:
+        nodeUpgradeDateTime = dateutil.parser.parse(upgradeSchedule[nodeId])
+        nodeUpgradeDateTime += timedelta(days=5)
+        upgradeSchedule[nodeId] = nodeUpgradeDateTime.isoformat()
+    return _validUpgrade
+
+
+def testPoolUpgradeScheduledOnProperDate(poolNodesStarted,
```
ashcherbakov: Please use snake case

indy_client/test/cli/test_pool_upgrade_schedule.py
```diff
@@ -0,0 +1,26 @@
+import dateutil
```
ashcherbakov: I think this test needs to be in indy_node.test.upgrade

spivachuk: This test operates on CLI level (depends on `poolUpgradeSubmitted` fixture). So it should be located in `indy_client` package.

PR #394, opened 2017-10-11
INDY-157: Fixed bug with wrong cancel messages

spivachuk:
- Fixed a bug with wrong "when" item in Upgrader.scheduledUpgrade tuple. This in turn fixed the bug with wrong "when" field in "cancel" messages of the upgrade log.
- Corrected tests according to the fix.
- Made some minor fixes (in tests) related to appending messages to the upgrade log.


indy_node/test/upgrade/helper.py
```diff
@@ -133,7 +133,9 @@ def check_node_set_acknowledges_upgrade(
         node_ids=node_ids)
 
     for node in node_set:
-        node.upgrader.scheduledUpgrade = (version, 0, randomString(10))
+        node.upgrader.scheduledUpgrade = (version,
+                                          datetime.utcnow().replace(tzinfo=dateutil.tz.tzutc()),
```
ashcherbakov: We can use `get_utc_epoch()` utility method for the same purpose

spivachuk: `get_utc_epoch()` returns UTC timestamp (in seconds as `int` value) rather than `datetime`.

PR #405, opened 2017-10-18
INDY-833: Init bls stewards

ashcherbakov:
- Add a script to enable BLS for Steward
- Improved NODE txn tests


indy_client/test/cli/test_node.py
```diff
@@ -7,59 +7,89 @@
...
+                                     newStewardCli, newNodeVals,
+                                     newNodeAdded):
     be(newStewardCli)
     newNodeVals['newNodeData'][ALIAS] = randomString(6)
     doSendNodeCmd(do, newNodeVals,
                   expMsgs=['existing data has conflicts with request data'])
     exitFromCli(do)
+
+
+def test_update_alias(be, do,
```
mzk-vct: Please add description for this test

setup.py
```diff
@@ -119,7 +119,8 @@ def run(self):
              'scripts/test_some_write_keys_others_read_them',
              'scripts/test_users_write_and_read_own_keys',
              'scripts/validator-info',
-             'scripts/init_bls_keys'],
+             'scripts/init_bls_keys',
+             'scripts/init_bls_for_stewards'],
```
mzk-vct: This name is confusing.
As far as I understand this script creates bls keys and sends transaction to publish them, but it's name tells that it adds bls keys for steward (as if steward can have bls keys)

scripts/init_bls_for_stewards
```diff
@@ -0,0 +1,117 @@
...
+
+
+def parse_args():
+    parser = argparse.ArgumentParser(
+        description="Generate BLS keys for a node "
+                    "by taking the node's name and seeds "
+                    "and send NODE txn with the BLS key specified")
+
+    parser.add_argument('--name', required=True, help='node name')
+    parser.add_argument('--node_seed', required=True, type=str,
```
mzk-vct: If node_seed is used only for generation of `dest` could you please add option to specify `dest` directly?

ashcherbakov: The requirements for the task is that BLS key needs to be created for a seed (the same seed as used for node keys actually). So I think this option is not required (let's add it if this requirement appears)

setup.py
ashcherbakov: Renamed to `enable_bls`

indy_client/test/cli/test_node.py
ashcherbakov: Done

PR #408, opened 2017-10-20
INDY-382: Fixed bug with forced config txns

spivachuk:
- Fixed a bug with repeated attempt to handle forced config transactions.
- Corrected UpgradeLog.__iter__ method.
- Added a test verifying that node upgrade is not scheduled twice when pool upgrade is done in force mode.
- Added tests verifying actual schedule of pool upgrade for the cases when POOL_UPGRADE transaction is sent once and twice.
- Made an assertion more exact in checkUpgradeScheduled test helper function.
- Isolated node upgrade tests from each other (there was interference between them previously).
- Corrected names of some tests.


indy_client/test/cli/test_pool_upgrade_schedule.py
```diff
@@ -19,7 +19,7 @@ def validUpgrade(_validUpgrade):
     return _validUpgrade
 
 
-def testPoolUpgradeScheduledOnProperDate(poolNodesStarted,
+def testNodeUpgradeScheduledOnProperDate(poolNodesStarted,
```
ashcherbakov: Why did we rename it? We test POOL_UPGRADE here, don't we?

indy_node/test/upgrade/test_forced_upgrade_no_consensus_on_single_node.py
```diff
@@ -0,0 +1,27 @@
...
+            node.stop()
+        else:
+            node.upgrader.scheduledUpgrade = None
+    sendUpgrade(trustee, trusteeWallet, nup)
+
+    def testsched():
+        for node in nodeSet:
+            if node.name == "Alpha":
+                assert node.upgrader.scheduledUpgrade
+                assert node.upgrader.scheduledUpgrade[0] == nup[VERSION]
```
ashcherbakov: Didn't we bump the version, so it should be more than nup[VERSION]?

indy_client/test/cli/test_pool_upgrade_schedule.py
spivachuk: The module name is `test_pool_upgrade_schedule` since we test the pool upgrade schedule here.
For each node upgrade is scheduled on individual date/time. The test verifies that for each node the upgrade is scheduled on the proper date, so its name is `testNodeUpgradeScheduledOnProperDate`.

indy_node/test/upgrade/test_node_schedules_upgrade_exp_force_false.py
```diff
@@ -0,0 +1,2 @@
+def testNodeSchedulesUpgradeExpForceFalse(upgradeScheduledExpForceFalse):
```
ashcherbakov: Please use snake case

indy_node/test/upgrade/test_node_schedules_upgrade_exp_force_true.py
```diff
@@ -0,0 +1,2 @@
+def testNodeSchedulesUpgradeExpForceTrue(upgradeScheduledExpForceTrue):
```
ashcherbakov: Please use snake case for test name

indy_node/test/upgrade/test_node_schedules_upgrade.py
```diff
@@ -0,0 +1,2 @@
+def testNodeSchedulesUpgrade(upgradeScheduled):
```
ashcherbakov: Please use snake case for test name

indy_node/test/upgrade/test_pool_upgrade_force_scheduled_only_once.py
```diff
@@ -0,0 +1,9 @@
+from indy_node.server.upgrade_log import UpgradeLog
+
+
+def testPoolUpgradeForceScheduledOnlyOnce(upgradeScheduledExpForceTrue,
```
ashcherbakov: Please use snake case for test name

indy_client/test/cli/test_pool_upgrade_schedule.py
ashcherbakov: Please use snake case for test name

indy_node/test/upgrade/test_forced_upgrade_no_consensus_on_single_node.py
spivachuk: This test was written earlier. In scope of this pull request it is only moved to a separate file.
As I can see, we verify that the target version of the scheduled upgrade is the same as we send in `sendUpgrade` call.

indy_node/test/upgrade/test_pool_upgrade_force_scheduled_only_once.py
ashcherbakov: Do we have a test that schedules update twice and checks that they work as expected (as in INDY-382)?

andkononykhin:
test this please

andkononykhin:
test this please

indy_node/test/upgrade/test_pool_upgrade_force_scheduled_only_once.py
spivachuk: Added the tests `test_node_schedules_upgrade_for_proper_datetime` and `test_node_reschedules_upgrade_for_proper_datetime` for verifying actual schedule of pool upgrade for the cases when POOL_UPGRADE transaction is sent once and twice.

PR #437, opened 2017-11-07
Fix migration script

dsurnin:


data/migrations/deb/1_2_188_to_1_2_189.py
```diff
@@ -76,110 +71,117 @@ def get_network_name():
...
-    # Move wallets directory
-    logger.info('Move wallets directory {} -> {}'.format(
-        old_dir, new_dir))
-    shutil.move(old_dir, new_dir)
-    logger.info('done')
-
-    # Move txns for CLI
-    migrate_genesis_txn(old_dir, new_txn_dir)
+def remove_network_from_dir_name(root_dir):
+    logger.error("remove_network_from_dir_name {}".format(root_dir))
```
sergey-shilov: May be logger.info() instead of logger.error()?

PR #455, opened 2017-11-16
Fix tests for NodeControlTool (#940)

anikitinDSR:


indy_node/test/upgrade/helper.py
```diff
@@ -1,3 +1,18 @@
+#   Copyright 2017 Sovrin Foundation
```
ashcherbakov: Please remove copyright header for now

PR #481, opened 2017-12-08
INDY-960: Rejected and later re-applied batches

spivachuk:
- Added a test verifying that a batch rejected due to catch-up start can be successfully re-applied and ordered later before ledgers synchronization (without the issue with lack of the batch in IdrCache.unCommitted list).
- Updated indy-plenum dependency.


indy_node/test/catchup/test_3pc_batch_reverted_and_later_ordered_on_catchup.py
```diff
@@ -0,0 +1,132 @@
...
+    Node.try_processing_ordered(self, msg)
+
+    if msg.instId == 0:
+        self.nodeIbStasher.reset_delays_and_process_delayeds()
+
+        self.try_processing_ordered = \
+            MethodType(Node.try_processing_ordered, self)
+
+
+def send_random_requests(wallet: Wallet, client: Client, count: int):
```
ashcherbakov: Why can't we use a helper method from plenum?

indy_node/test/catchup/test_3pc_batch_reverted_and_later_ordered_on_catchup.py
```diff
@@ -0,0 +1,132 @@
...
+
+    Batched._should_batch = original_should_batch
+
+
+concerningLogLevels = [logging.WARNING,
+                       logging.ERROR,
+                       logging.CRITICAL]
+
+
+def test_3pc_batch_reverted_on_catchup_start_can_be_ordered_before_ledgers_sync(
```
ashcherbakov: Please add a description on what this test does

PR #486, opened 2017-12-11
Config ledger plenum

lovesh:


indy_node/server/client_authn.py
```diff
@@ -18,23 +15,10 @@ class LedgerBasedAuthNr(CoreAuthMixin, NaclAuthNr):
...
-    query_types = CoreAuthMixin.query_types.union(
-        {GET_TXNS, }
-    ).union(
-        PoolRequestHandler.query_types
-    ).union(
-        DomainReqHandler.query_types
-    ).union(
-        ConfigReqHandler.query_types
-    )
+    write_types = CoreAuthMixin.write_types.union({ATTRIB, SCHEMA, CLAIM_DEF,
```
ashcherbakov: Why do we have to specify it explicitly? Why can't get it from registered request handlers?

indy_node/server/node.py
```diff
@@ -93,21 +88,6 @@ def __init__(self,
...
-            postCatchupCompleteClbk=self.postConfigLedgerCaughtUp,
-            postTxnAddedToLedgerClbk=self.postTxnFromCatchupAddedToLedger)
-        self.on_new_ledger_added(CONFIG_LEDGER_ID)
-        self.states[CONFIG_LEDGER_ID] = self.loadConfigState()
-        self.upgrader = self.getUpgrader()
-        self.poolCfg = self.getPoolConfig()
-        self.configReqHandler = self.getConfigReqHandler()
-        self.initConfigState()
-        self.register_req_handler(CONFIG_LEDGER_ID, self.configReqHandler)
-        self.requestExecuter[CONFIG_LEDGER_ID] = self.executeConfigTxns
```
ashcherbakov: Where do we register executors for CONFIG ledger?

indy_node/server/node.py
```diff
@@ -378,15 +302,8 @@ def executeDomainTxns(self, ppTime, reqs: List[Request], stateRoot,
...
-        req_handler = self.get_req_handler(DOMAIN_LEDGER_ID)
-        return self.commitAndSendReplies(req_handler, ppTime, reqs,
-                                         stateRoot, txnRoot)
-
-    def executeConfigTxns(self, ppTime, reqs: List[Request], stateRoot,
-                          txnRoot) -> List:
-        req_handler = self.get_req_handler(CONFIG_LEDGER_ID)
-        return self.commitAndSendReplies(req_handler, ppTime, reqs,
-                                         stateRoot, txnRoot)
+        return self.default_executer(DOMAIN_LEDGER_ID, ppTime, reqs,
```
ashcherbakov: Why can't we have a generic method in plenum for execution of txns?
Do we need to override `executeDomainTxns`?

indy_node/server/node.py
lovesh: Yes

indy_node/server/client_authn.py
lovesh: Because request handlers are registered with node not authenticator.

indy_node/server/node.py
lovesh: Because `executeDomainTxns` in plenum does an extra action which is not need for this one

PR #487, opened 2017-12-11
INDY-878: add migration script, cleanup config, increase plenum version.

sergey-shilov:


data/migrations/deb/1_2_233_to_1_2_234.py
```diff
@@ -0,0 +1,40 @@
...
+import indy_node.general_config.indy_config as indy_config
+
+
+def migrate():
+    config = getConfig()
+    logger = getlogger()
+
+    general_config_path = os.path.join(config.GENERAL_CONFIG_DIR, config.GENERAL_CONFIG_FILE)
+
+    if os.path.exists(general_config_path):
```
ashcherbakov: That's not a big problem for migration script and this method, but in general I prefer not having too many embedded conditions, and rather use 
```
if not os.path.exists(general_config_path):
    return
```

sergey-shilov: Ok

PR #506, opened 2017-12-22
Script for export DIDs from CLI wallet

jovfer:


setup.py
```diff
@@ -93,5 +93,6 @@
              'scripts/validator-info',
              'scripts/init_bls_keys',
              'scripts/enable_bls',
-             'scripts/create_dirs.sh']
+             'scripts/create_dirs.sh',
+             'scripts/indy_export_dids']
```
ashcherbakov: Probably we can rename the script to `indy_old_cli_export_dids`?

scripts/indy_export_dids
```diff
@@ -0,0 +1,50 @@
...
+    seed_base64 = base64.b64encode(did_signer.seed).decode("ascii")
+    dids.append({"did": did, "seed": seed_base64})
+
+dto = {
+    "version": 1,
+    "dids": dids
+}
+
+out_file_name = "{}_{}.exp_wallet".format(env_name, wallet_name)
+path = Path(os.path.join(os.path.curdir, out_file_name))
```
ashcherbakov: Should we provide an option to set the destination folder for the exported data?

scripts/indy_export_dids
```diff
@@ -0,0 +1,50 @@
...
+
+wf = path.open('w')
+wf.write(json.dumps(dto))
```
ashcherbakov: We can use `write_text` here to open, write and close the file.

scripts/indy_export_dids
jovfer: Or may be even out file name? Default destination folder = current folder, and optional argument - path to save file (including filename).

PR #520, opened 2018-01-12
INDY-1087: add iptables rules to limit the number of clients connections.

sergey-shilov:


setup.py
```diff
@@ -94,5 +94,6 @@
              'scripts/init_bls_keys',
              'scripts/enable_bls',
              'scripts/create_dirs.sh',
-             'scripts/indy_old_cli_export_dids']
+             'scripts/indy_old_cli_export_dids',
+             'scripts/setup_indy_node_iptable']
```
ashcherbakov: missing 's' at the end of file name

PR #522, opened 2018-01-15
INDY-1083: do not allow overriding of existing claim defs

ashcherbakov:


indy_node/server/domain_req_handler.py
```diff
@@ -214,6 +200,40 @@ def _validateAttrib(self, req: Request):
...
+                                       'name {} and version {}'
+                                       .format(identifier, schema_name, schema_version))
+
+    def _validate_claim_def(self, req: Request):
+        # we can not add a Claim Def with existent ISSUER_DID
+        # sine a Claim Def needs to be identified by seqNo
+        identifier = req.identifier
+        operation = req.operation
+        schema_ref = operation[REF]
+        claim_def, _, _, _ = self.getClaimDef(
```
lovesh: An issuer should be allowed to have multiple claim definitions for the same schema, it's just that they need to be for different signature schemes

indy_node/server/domain_req_handler.py
```diff
@@ -214,6 +200,40 @@ def _validateAttrib(self, req: Request):
...
+            author=identifier,
+            schemaSeqNo=schema_ref
+        )
+        if claim_def:
+            raise InvalidClientRequest(identifier, req.reqId,
+                                       '{} can have one and only one CLAIM_DEF with '
+                                       'issuer DID (identifier) {} and schema ref {}'
```
lovesh: This should read 'signature scheme {} and schema ref {}'

indy_node/server/domain_req_handler.py
ashcherbakov: Done

indy_node/server/domain_req_handler.py
ashcherbakov: Done

PR #531, opened 2018-01-22
INDY-1069: move state-depended checks from static to dynamic validation.

sergey-shilov:

### DIFF
--- a/indy_client/test/cli/test_nym.py
+++ b/indy_client/test/cli/test_nym.py
@@ -1,3 +1,4 @@
+import json
 import pytest
 from plenum.common.signer_did import DidSigner
 from plenum.common.signer_simple import SimpleSigner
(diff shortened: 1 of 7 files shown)


indy_client/test/cli/test_nym.py
```diff
@@ -361,3 +364,18 @@ def test_send_different_nyms_succeeds_when_batched(
 
     do('send GET_NYM dest={dest}',
         mapper=parameters, expect=CURRENT_VERKEY_FOR_NYM, within=2)
+
+
+def test_nym_resend(looper, sdk_pool_handle, sdk_wallet_steward):
```
ashcherbakov: This test should not be in the CLI folder

indy_node/server/domain_req_handler.py
```diff
@@ -142,6 +138,10 @@ def _validateNym(self, req: Request):
         origin = req.identifier
         op = req.operation
 
+        s, reason = self.canNymRequestBeProcessed(origin, op)
```
ashcherbakov: Do we need this call at all? Doesn't `_validateExistingNym` already do similar validation based on `Authoriser`?

indy_client/test/cli/test_nym.py
sergey-shilov: Ok, moved to separate file.

indy_node/server/domain_req_handler.py
sergey-shilov: Yeah, seems like it is double-check, have removed.

sergey-shilov:
Test this please

PR #536, opened 2018-01-25
INDY-886: proposal to change the format of transactions and requests 

ashcherbakov:


hyperledger-github:

ashcherbakov:
@swcurran Some types were made as enumerated integers to support possible renaming of, for example, transactions (for example, from CLAIM_DEF to CRED_DEF).

docs/requests-new.md
```diff
@@ -0,0 +1,1941 @@
...
+            "auditProof": <...>, 
+        ]
+    },
+    "metadata": {
+        "reqId": <...>,
+        "from": <...>,
+    },
+}
+```
+where `multiSignature`'s `signedState` is a serialized value having the following form:
```
lovesh: @AlexanderShekhovcov Where is the actual serialised data that was signed? Or does the sdk know how to serialise the data

ashcherbakov: This is the data serialized and signed by nodes. So, the serialized data is signed, not actual one.
This data is not re-serialized on nodes side and passed to clients in serialized format. So, the client can verify the signature over serialized data.
Then it can de-serialize the data according to the serialization specified in "serialization" field.

BTW the same approach is going to be applied for Requests, where signature is calculated against serialized data, and the serialized data is passed to the pool, not actual one, and re-serializations on node side for requests is not needed.

lovesh: But what is the data that is serialised? `signedState` just has 2 metadata objects, eg. in a `GET_NYM` txn, the serialised data will have the actual NYM and verkey, does that not live under `signedState`?

lovesh: Ok, can you point out where is the data that is serialised for verifying the proof?

ashcherbakov: Data used to verify BLS multisig is in `signedState `
Data used to verify state proof (I think you mean the real result for the get request including key and value in state trie) is in <result>.

docs/transactions-new.md
```diff
@@ -0,0 +1,749 @@
...
+transactions defining the Pool and the Network.
+- pool genesis transactions defining initial trusted Nodes in the Pool
+- domain genesis transactions defining initial trusted Trustees and Stewards
+
+## Common Structure
+Each transaction has the following structure containing of metadata values (common for all transaction types) and 
+transaction specific data:
+```
+{
+    "txn": {
```
lovesh: Its not functionally wrong, but `txn` under transaction seems odd, maybe call it `payload`

docs/transactions-new.md
```diff
@@ -0,0 +1,749 @@
...
+    "txnMetadata": {
+        "ver": <...>,
+        "creationTime": <...>,
+        "seqNo": <...>,  
+    },
+    "reqSignature": {
+        "type": <...>,
+        "ver": <...>,
+        "values": [
+            "from": <...>,
```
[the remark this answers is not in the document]
kdenhartog: A submission could be submitted by a trust anchor on behalf of a DID owner in which case knowing who submitted the transaction and which DID the transaction is about.

docs/transactions-new.md
```diff
@@ -0,0 +1,749 @@
...
+    },
+    "txnMetadata": {
+        "ver": <...>,
+        "creationTime": <...>,
+        "seqNo": <...>,  
+    },
+    "reqSignature": {
+        "type": <...>,
+        "ver": <...>,
+        "values": [
```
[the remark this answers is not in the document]
kdenhartog: I agree with the sentiment of needing to modify the "values" key to an array such that it can accept a multisig. Would we also need to move ver (I think this means verkey) into this object to support this new format? Also, symantically I'd prefer that we refer to them as signatures instead of values, but it's not incredibly important because it's stated above it.

docs/transactions-new.md
kdenhartog: +1 to switching this to payload, but again comes down to semantics, so it's more of a preference than a necessity.

docs/transactions-new.md
```diff
@@ -0,0 +1,749 @@
...
+#### NYM
+Creates a new NYM record for a specific user, trust anchor, steward or trustee.
+Note that only trustees and stewards can create new trust anchors and trustee can be created only by other trusties (see [roles](https://docs.google.com/spreadsheets/d/1TWXF7NtBjSOaUIBeIH77SyZnawfo91cJ_ns4TR-wsq4/edit#gid=0)).
+
+The transaction can be used for 
+creation of new DIDs, setting and rotation of verification key, setting and changing of roles.
+
+- `id` (string):
+
+    Nym's ID as State Trie key (address or descriptive data). It must be unique within the ledger.
```
lovesh: Can you give an example of `id`? If `did1` is creating (`did1` is a trust anchor) `did2`, what is `id`? Hash of `did1` or hash of `did2`? Also when do you need to look at a NYM in ledger txn and then go to state trie? And if you need state key, do we need to store the key, cant we store a reference to a mechanism like `1->sha256`, `2->sha3`, etc

docs/requests-new.md
ashcherbakov: Sorry, in <result>

docs/transactions-new.md
ashcherbakov: Yes, `payload` is probably better.

docs/transactions-new.md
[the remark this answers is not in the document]
ashcherbakov: Yes, it's assumed to be as Lovesh provided, just a typo in the doc.

docs/transactions-new.md
ashcherbakov: We should use a general `txnId` instead of `id`. It can be used to uniquely identify a txn in the state. So, it can contain a key in the state trie.

PR #543, opened 2018-01-30
Removing apt-mark usage on non Debian distros

M0Rf30:
This condition allow to use node_control_tool.py with an agnostic approach towards Linux distros

### DIFF
--- a/indy_node/utils/node_control_tool.py
+++ b/indy_node/utils/node_control_tool.py
@@ -99,17 +99,20 @@ class NodeControlTool:
         return ret.stdout.strip()
 
     def _hold_packages(self):
-        cmd = compose_cmd(['apt-mark', 'hold', self.packages_to_hold])
-        ret = self._run_shell_command(cmd)
-        if ret.returncode != 0:
-            raise Exception('cannot mark {} packages for hold '
-                            'since {} returned {}'
-                            .format(self.packages_to_hold,
-                                    cmd,
-                                    ret.returncode))
+        if shutil.which("apt-mark"):
+            cmd = compose_cmd(['apt-mark', 'hold', self.packages_to_hold])
+            ret = self._run_shell_command(cmd)
+            if ret.returncode != 0:
+                raise Exception('cannot mark {} packages for hold '
+                                'since {} returned {}'
+                                .format(self.packages_to_hold,
+                                        cmd,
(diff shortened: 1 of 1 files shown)


ashcherbakov:
test this please

M0Rf30:
I've tested on Arch Linux and it works. You should test on ubuntu

ashcherbakov:
@M0Rf30 
Sorry for confusion, `test this please` was a command to CI (Jenkins) to start integration tests (it's a required step to merge the PR).

hyperledger-github:

indy_node/utils/node_control_tool.py
```diff
@@ -272,7 +273,8 @@ def _run_shell_script(cls, command, timeout):
                               timeout=timeout)
 
     def start(self):
-        self._hold_packages()
+        if 'ubuntu' in platform.platform().lower():
```
andkononykhin: why 'ubuntu' only, e.g. such a conditional will fail on debian where apt is also available

M0Rf30: So checking of apt-mark executable is needed to be more generic

M0Rf30:
I've modified the commit. Code checks for apt-mark presence, so it should work on every distro

M0Rf30:
test this please

indy_node/utils/node_control_tool.py
```diff
@@ -271,7 +271,8 @@ def _run_shell_script(cls, command, timeout):
                               timeout=timeout)
 
     def start(self):
-        self._hold_packages()
+        if shutil.which("apt-mark"):
+            self._hold_packages()
```
andkononykhin: It will work. But I don't think that it is to skip hold-packages stage silently. Also if we someday want to call `_hold_packages` from some other place we should copy-paste the conditional you added. Does it make sense to move the conditional inside the method itself? Could you add some logging/warning in case no operations performed?

M0Rf30:
Now should be OK

indy_node/utils/node_control_tool.py
andkononykhin: Please, add logging (warning) for the case when we skip packages holding. Thank you

andkononykhin:
test this please

PR #548, opened 2018-02-01
INDY-1074: add ATTRIB hash and enc support, fix and unskip tests.

sergey-shilov:

### DIFF
--- a/indy_common/state/domain.py
+++ b/indy_common/state/domain.py
@@ -205,8 +205,9 @@ def prepare_get_attr_for_state(txn):
     if attr_type == ENC:
         attr_key = hash_of(attr_key)
 
-    path = make_state_path_for_attr(nym, attr_key)
-    return path, None, None, None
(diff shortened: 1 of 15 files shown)


hyperledger-github:

indy_common/state/domain.py
```diff
@@ -75,13 +75,17 @@ def prepare_attr_for_state(txn):
     """
     assert txn[TXN_TYPE] in {ATTRIB, GET_ATTR}
     nym = txn[TARGET_NYM]
-    attr_key, value = parse_attr_txn(txn)
-    hashed_value = hash_of(value) if value else ''
+    attr_type, attr_key, value = parse_attr_txn(txn)
+    if attr_type == ENC:
```
ashcherbakov: Can't we just remove 
```
    if attr_type == ENC:
        return hash_of(attr), attr
```
from `parse_attr_txn`?

sergey-shilov: Agreed, done.

PR #550, opened 2018-02-02
[INDY-1109] Added NETWORK_NAME into config after install indy-node

anikitinDSR:


hyperledger-github:

PR #555, opened 2018-02-06
INDY-582: support setting of empty (None) verkey on client and server sides

sergey-shilov:
Now any NYM owner can flush its verkey. In this case permissions for verkey updating goes back to creator.


hyperledger-github:

PR #557, opened 2018-02-08
INDY-1081: Updated indy-plenum dependency

spivachuk:


hyperledger-github:

PR #568, opened 2018-02-18
INDY-1181: Suggest amd64 arch for packer build / fix box name

beatchristen:


hyperledger-github:

beatchristen:
replaced with #574 where Sign-Off header has been added

PR #576, opened 2018-02-21
REVOC_DEF_ENTRY support

anikitinDSR:


hyperledger-github:

indy_common/state/domain.py
```diff
@@ -135,6 +144,22 @@ def prepare_revoc_def_for_state(txn):
...
+    assert author_did
+    assert revoc_reg_def_id
+    path = make_state_path_for_revoc_reg_entry(authors_did=author_did,
+                                               revoc_reg_def_id=revoc_reg_def_id)
+
+    seq_no = txn[f.SEQ_NO.nm]
+    txn_time = txn[TXN_TIME]
+    assert seq_no
+    assert txn_time
+    value_bytes = encode_state_value(txn, seq_no, txn_time)
```
ashcherbakov: The state should accumulate all issued and revoked indices.
Please also add a test for this.

indy_node/server/domain_req_handler.py
```diff
@@ -252,6 +257,12 @@ def _validate_revoc_reg_def(self, req: Request):
         assert revoc_def_tag
         assert revoc_def_type
 
+    def _validate_revoc_reg_entry(self, req: Request):
+        operation = req.operation
+        revoc_reg_def_id = operation.get(REVOC_REG_DEF_ID)
+        assert revoc_reg_def_id
+        # TODO Add check that sufficient revoc_reg_def exist
```
ashcherbakov: Can we add this check now?

indy_common/types.py
```diff
@@ -106,6 +106,23 @@ class ClientRevocDefSubmitField(MessageValidator):
     )
 
 
+class RevocRegEntryValueField(MessageValidator):
+    schema = (
+        (PREV_ACCUM, NonEmptyStringField()),
+        (ACCUM, NonEmptyStringField()),
+        (ISSUED, IterableField(inner_field_type=IntegerField())),
+        (REVOKED, IterableField(inner_field_type=NonEmptyStringField()))
```
ashcherbakov: Should it be `IntegerField` here as well?

indy_node/server/domain_req_handler.py
```diff
@@ -22,9 +23,172 @@
 logger = getlogger()
 
 
+class RevocationStrategy:
```
ashcherbakov: Please move the strategies into a separate file

indy_node/server/domain_req_handler.py
```diff
@@ -22,9 +23,172 @@
...
+        self.author_did = None
+        self.revoc_reg_def_id = None
+        self.req_id = None
+
+    def set_parameters_from_txn(self, author_did, revoc_reg_def_id, req_id):
+        self.author_did = author_did
+        self.revoc_reg_def_id = revoc_reg_def_id
+        self.req_id = req_id
+
+    def set_strategy(self, strategy):
```
ashcherbakov: Why do we need this? We use inheritance for this, don't we? So, it's better to use abstract method to perform strategy-specific validation (do not forget to use @abstractmethod for an abstract class)

indy_node/server/domain_req_handler.py
ashcherbakov: I believe it should be an abstract class with `metaclass=ABCMeta` inheritance.

indy_node/server/domain_req_handler.py
```diff
@@ -22,9 +23,172 @@
...
+        # General checks for all Revocation entries
+        operation = req.operation
+        value_from_txn = operation.get(VALUE)
+        issued_from_txn = value_from_txn.get(ISSUED)
+        revoked_from_txn = value_from_txn.get(REVOKED)
+        intersection = set(issued_from_txn).intersection(set(revoked_from_txn))
+        if len(intersection) > 0:
+            raise InvalidClientRequest(self.author_did,
+                                       self.req_id,
+                                       "Indices {} are existed in issued and revoked list".format(intersection))
```
ashcherbakov: Probably it's better to say: 'Can not have an index in both "issued" and "revoked" lists'.

indy_node/server/domain_req_handler.py
```diff
@@ -22,9 +23,172 @@
...
+        if current_entry is None:
+            return None
+        value_from_state = current_entry.get(VALUE)
+        assert value_from_state
+        current_accum = value_from_state.get(ACCUM)
+        if current_accum != value_from_txn.get(PREV_ACCUM):
+            raise InvalidClientRequest(self.author_did,
+                                       self.req_id,
+                                       "Accum value from state: {} "
+                                       "do not equal with {} value "
```
ashcherbakov: It's better to say:
The current accumulator value {} must be equal to the last accumulator value {} in transaction.

indy_node/server/domain_req_handler.py
```diff
@@ -22,9 +23,172 @@
...
+                                       "from txn: {}".format(current_accum,
+                                                             PREV_ACCUM,
+                                                             value_from_state.get(PREV_ACCUM)))
+
+        # Strategy specific validation
+        self.strategy.validate(current_entry, req)
+
+    def write(self, current_reg_entry, txn):
```
ashcherbakov: Use @abstractmethod

indy_node/server/domain_req_handler.py
```diff
@@ -22,9 +23,172 @@
...
+        indices = value_from_state.get(REVOKED, [])
+        value_from_txn = req.operation.get(VALUE)
+        issued_from_txn = value_from_txn.get(ISSUED, [])
+        revoked_from_txn = value_from_txn.get(REVOKED, [])
+        issued_difference = set(issued_from_txn).difference(indices)
+        if len(issued_difference) > 0:
+            raise InvalidClientRequest(self.author_did,
+                                       self.req_id,
+                                       "Issued indices from txn: {} "
+                                       "does not exist in current "
```
ashcherbakov: better to say: are not present in the current ....

indy_node/server/domain_req_handler.py
```diff
@@ -22,9 +23,172 @@
...
+        indices = value_from_state.get(ISSUED, [])
+        value_from_txn = req.operation.get(VALUE)
+        issued_from_txn = value_from_txn.get(ISSUED, [])
+        revoked_from_txn = value_from_txn.get(REVOKED, [])
+        revoked_difference = set(revoked_from_txn).difference(indices)
+        if len(revoked_difference) > 0:
+            raise InvalidClientRequest(self.author_did,
+                                       self.req_id,
+                                       "Revoked indices from txn: {} "
+                                       "does not exist in current "
```
ashcherbakov: better to say: are not present in the current ....

indy_node/server/domain_req_handler.py
```diff
@@ -22,9 +23,172 @@
...
+                                       "issued list from state: {}".format(revoked_difference,
+                                                                           indices))
+        issued_intersection = set(indices).intersection(issued_from_txn)
+        if len(issued_intersection) > 0:
+            raise InvalidClientRequest(self.author_did,
+                                       self.req_id,
+                                       "Issued indices from txn: {} "
+                                       "already issued "
```
ashcherbakov: are already...

indy_node/server/domain_req_handler.py
```diff
@@ -22,9 +23,172 @@
...
+            value_from_state = current_reg_entry.get(VALUE)
+            assert value_from_state
+            indices = value_from_state.get(ISSUED, [])
+            value_from_txn = txn.get(VALUE)
+            issued_from_txn = value_from_txn.get(ISSUED, [])
+            revoked_from_txn = value_from_txn.get(REVOKED, [])
+            # set with all previous issued minus revoked from txn
+            result_indicies = set(indices).difference(revoked_from_txn)
+            result_indicies.update(issued_from_txn)
+            value_from_txn[ISSUED] = []
```
ashcherbakov: Shouldn't we use ISSUED value here?

indy_node/server/domain_req_handler.py
```diff
@@ -22,9 +23,172 @@
...
+        path, value_bytes = domain.prepare_revoc_reg_entry_for_state(txn)
+        self.state.set(path, value_bytes)
+
+
 class DomainReqHandler(PHandler):
-    write_types = {NYM, ATTRIB, SCHEMA, CLAIM_DEF, REVOC_REG_DEF}
+    write_types = {NYM, ATTRIB, SCHEMA, CLAIM_DEF, REVOC_REG_DEF, REVOC_REG_ENTRY}
     query_types = {GET_NYM, GET_ATTR, GET_SCHEMA, GET_CLAIM_DEF}
+    revocation_strategy_map = {
+        "ISSUANCE_BY_DEFAULT": RevokedStrategy,
```
ashcherbakov: Please use constants for "ISSUANCE_BY_DEFAULT" and "ISSUANCE_ON_DEMAND"

indy_node/server/domain_req_handler.py
```diff
@@ -252,6 +423,32 @@ def _validate_revoc_reg_def(self, req: Request):
         assert revoc_def_tag
         assert revoc_def_type
 
+    def get_current_revoc_entry_and_revoc_def(self, author_did, revoc_reg_def_id, req_id):
+        assert author_did
+        assert revoc_reg_def_id
+        # TODO Need to be "isCommited"==True?
```
ashcherbakov: Since we call this method as part of dynamic validation (that is for new txns added to uncommitted list), we must get the uncommitted values to take into account all txns in the batch

indy_node/server/domain_req_handler.py
```diff
@@ -252,6 +423,32 @@ def _validate_revoc_reg_def(self, req: Request):
...
+        current_entry, _, _, _ = self.getRevocDefEntry(author_did=author_did,
+                                                       revoc_reg_def_id=revoc_reg_def_id,
+                                                       isCommitted=False)
+        # TODO Need to be "isCommited"==True ?
```
ashcherbakov: The same as above

indy_node/server/domain_req_handler.py
```diff
@@ -252,6 +423,32 @@ def _validate_revoc_reg_def(self, req: Request):
         assert revoc_def_tag
         assert revoc_def_type
 
+    def get_current_revoc_entry_and_revoc_def(self, author_did, revoc_reg_def_id, req_id):
```
ashcherbakov: Should it be a private method? (with _ prefix)

indy_node/server/domain_req_handler.py
```diff
@@ -252,6 +423,32 @@ def _validate_revoc_reg_def(self, req: Request):
...
+
+    def _validate_revoc_reg_entry(self, req: Request):
+        current_entry, revoc_def = self.get_current_revoc_entry_and_revoc_def(
+            author_did=req.identifier,
+            revoc_reg_def_id=req.operation[REVOC_REG_DEF_ID],
+            req_id=req.reqId
+        )
+        validator_cls = self.get_revocation_strategy(revoc_def[VALUE][ISSUANCE_TYPE])
+        validator = validator_cls(self.state)
+        validator.set_strategy(validator)
```
ashcherbakov: Ugly line! See comment about inheritance.

PR #591, opened 2018-03-05
INDY-1113: Corrected existing tests

spivachuk:
- Corrected existing tests according to the fixes in the procedures of node disconnection and reconnection.
- Updated indy-plenum-dev dependency.


hyperledger-github:

PR #593, opened 2018-03-06
INDY-1148 : update plenum version

Toktar:
update plenum version to 1.2.264


hyperledger-github:

PR #595, opened 2018-03-07
[INDY-1102] Added logging configure

anikitinDSR:


hyperledger-github:

PR #598, opened 2018-03-12
INDY-1111: prevent creating a schema or claimDef without role

Toktar:
- added validation for schema and claimDef adding
- expanded the permission map in auth.py
- added unit tests for auth.py
- added integration tests for domain_req_handler.py

### DIFF
--- a/indy_client/test/anon_creds/test_claim_def.py
+++ b/indy_client/test/anon_creds/test_claim_def.py
@@ -8,6 +8,7 @@ from stp_core.common.log import getlogger
 
 logger = getlogger()
 
+
 def test_error_submit_claim_def_by_client(submitted_schema_ID,
                                           public_key,
                                           public_revocation_key,
@@ -99,4 +100,3 @@ def test_submit_claim_def_same_schema_and_signature_type(submitted_claim_def,
                                             pk=public_key,
                                             pkR=public_revocation_key,
                                             signatureType='CL'))
-
--- a/indy_client/test/anon_creds/test_claim_def.py
+++ b/indy_client/test/anon_creds/test_claim_def.py
@@ -8,22 +8,21 @@ from stp_core.common.log import getlogger
 
 logger = getlogger()
 
+def test_error_submit_claim_def_by_client(submitted_schema_ID,
+                                          public_key,
+                                          public_revocation_key,
+                                          looper,
+                                          public_repo_for_client):
+    with pytest.raises(OperationError) as ex_info:
+        looper.run(public_repo_for_client.submitPublicKeys(id=submitted_schema_ID,
+                                                  pk=public_key,
+                                                  pkR=public_revocation_key,
+                                                  signatureType='CL'))
+        assert "role cannot add claim def" in ex_info[0]
+
 
 def test_submit_claim_def(submitted_claim_def):
     assert submitted_claim_def
-
-
-def test_submit_claim_def_same_schema_and_signature_type(submitted_claim_def,
-                                                         looper, public_repo,
-                                                         submitted_schema_ID,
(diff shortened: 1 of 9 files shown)


indy_client/test/anon_creds/test_claim_def.py
```diff
@@ -9,21 +9,21 @@
 logger = getlogger()
 
 
-def test_submit_claim_def(submitted_claim_def):
-    assert submitted_claim_def
+def test_error_submit_claim_def_by_client(submitted_schema_ID,
```
ashcherbakov: It's better to name it `test_can_not_submit_claim_def_by_identity_owner`.

indy_node/server/domain_req_handler.py
```diff
@@ -250,11 +270,38 @@ def _validate_claim_def(self, req: Request):
...
+            raise UnknownIdentifier(
+                req.identifier,
+                req.reqId)
         if claim_def:
-            raise InvalidClientRequest(identifier, req.reqId,
-                                       '{} can have one and only one CLAIM_DEF for '
-                                       'and schema ref {} and signature type {}'
-                                       .format(identifier, schema_ref, signature_type))
+            origin = req.identifier
+            owner = self.idrCache.getOwnerFor(req.identifier, isCommitted=False)
```
ashcherbakov: I think this check doesn't make sense. The owner is related to the DID who owns the corresponding verkey.
Here we just need to make sure that only the person who created the ClaimDef can modify it. But this is in fact always guaranteed since sender's DID (identifier) is part of the ClaimDef's key in the state trie.
So, I think that the whole check for whether we already have a Claim Def can be removed.

indy_client/test/anon_creds/test_schema.py
```diff
@@ -33,6 +33,15 @@ def test_submit_same_schema_twice(looper, public_repo,
         ex_info.match("can have one and only one SCHEMA with name GVT and version 1.0'")
 
 
+def test_submit_schema_without_role(looper, public_repo_for_client,
```
ashcherbakov: It's better to name it `test_can_not_submit_schema_by_identity_owner`.

indy_node/server/domain_req_handler.py
```diff
@@ -278,23 +278,14 @@ def _validate_claim_def(self, req: Request):
             raise UnknownIdentifier(
                 req.identifier,
                 req.reqId)
-        if claim_def:
```
ashcherbakov: There is no need to call `getClaimDef`, isn't it?

indy_common/auth.py
```diff
@@ -32,6 +32,10 @@ class Authoriser:
             {TRUSTEE: []},
         '{}_role_{}_'.format(NYM, TRUST_ANCHOR):
             {TRUSTEE: []},
+        '{}_role_<any>_<any>'.format(SCHEMA):
+            {TRUSTEE: [OWNER, ], STEWARD: [OWNER, ], TRUST_ANCHOR: [OWNER, ]},
```
spivachuk: An existing field of the transaction should be used rather than not existing (`role`). For `SCHEMA` transaction it makes sense to use `name` filed here.

indy_common/auth.py
```diff
@@ -32,6 +32,10 @@ class Authoriser:
...
+        '{}_role_<any>_<any>'.format(CLAIM_DEF):
+            {TRUSTEE: [OWNER, ], STEWARD: [OWNER, ], TRUST_ANCHOR: [OWNER, ]},
```
spivachuk: An existing field of the transaction should be used rather than not existing (`role`). For `CLAIM_DEF` transaction it makes sense to use `ref` filed here.

PR #599, opened 2018-03-13
INDY-1169 : Added static validation of schema

Toktar:
- Added validation of field 'attr_names'
- Added test for it


hyperledger-github:

PR #600, opened 2018-03-13
[INDY-1185] Fixed some failing tests in hyperledger jenkins

skhoroshavin:

### DIFF
--- a/indy_node/test/replay/test_successive_batch_no_state_change.py
+++ b/indy_node/test/replay/test_successive_batch_no_state_change.py
@@ -181,13 +181,15 @@ def test_successive_batch_do_no_change_state(looper,
 
     uncommitteds = {}
     methods = {}
+    allow_rejects = 0
     for node in nodeSet:
         cache = node.idrCache
         uncommitteds[cache._name] = []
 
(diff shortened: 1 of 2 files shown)


indy_node/test/replay/test_successive_batch_no_state_change.py
```diff
@@ -229,15 +230,17 @@ def patched_com(self, stateRoot):
...
-    for _ in range(3):
-        idy, _ = new_identity()
-        keys[idy.identifier] = idy.verkey
-        submit_id_req(idy)
-        looper.runFor(.01)
-
-    # Correct number of uncommitted entries
-    looper.run(eventually(check_uncommitted, 3, retryWait=1))
+    stashers = [n.nodeIbStasher for n in nodeSet]
+    with delay_rules(stashers, icDelay()):
```
ashcherbakov: Why do we prevent View Change only here? Don't we need to prevent the view change for the whole test?

skhoroshavin: @ashcherbakov in fact View Change was offending only this part (or at least it seemed so), so I tried to make minimal change in test behaviour

PR #602, opened 2018-03-13
[INDY-1138] Added storage for timestamp/root_hash revocation

anikitinDSR:


indy_common/config.py
```diff
@@ -51,6 +51,7 @@
 configStateDbName = 'config_state'
 attrDbName = 'attr_db'
 idrCacheDbName = 'idr_cache_db'
+tsRevocationDbName = "timestamp_revoc_db"
```
ashcherbakov: We need to store timestamps of all state updates (not only revocation) to reduce the risk of malicious nodes sending incorrect old results.
So, it's better to rename it to `state_ts_db`

indy_node/persistence/state_ts_store.py
```diff
@@ -0,0 +1,41 @@
...
+        value = self._storage.get(str(timestamp))
+        return value
+
+    def set(self, timestamp: int, root_hash: bytes):
+        self._storage.put(str(timestamp), root_hash)
+
+    def close(self):
+        self._storage.close()
+
+    def get_equal_or_prev(self, timestamp):
```
ashcherbakov: I think this needs to be moved to KeyValueStorage Interface level, so that we can easily replace leveldb by rockdsb in this class.

PR #605, opened 2018-03-16
INDY-1111: Refactoring

Toktar:
auth.py refactoring


hyperledger-github:

PR #610, opened 2018-03-19
[INDY-1137] GET_REVOC_REG and GET_REVOC_REG_DELTA

anikitinDSR:


indy_node/server/domain_req_handler.py
```diff
@@ -422,6 +426,122 @@ def handleGetRevocRegDefReq(self, request: Request):
...
+
+            reply_data = None if revoc_reg_entry is None else \
+                {
+                    REVOC_REG_DEF_ID: revoc_reg_def_id,
+                    REVOC_TYPE: revoc_reg_entry.get(REVOC_TYPE),
+                    VALUE: {
+                        ACCUM: revoc_reg_entry[VALUE].get(ACCUM)
+                    }
+                }
+            proof = self.make_proof(path, head_hash=past_root)
```
ashcherbakov: A separate path and entry needs to be used for GET_REVOC_REG in the state trie, otherwise the state proof can not be verified. The VALUE must be the same as the one stored in the state trie (here we store ACCUM+ISSUED/REVOKED, while return ACCUM only).
Also please add tests that state proofs in replies to GET_REVOC_REG and GET_REVOC_REG_DELTA can be verified.

indy_node/test/anon_creds/test_get_revoc_reg_delta_send_by_default.py
```diff
@@ -0,0 +1,90 @@
...
+    rev_reg_req2['operation'][VALUE][ACCUM] = randomString(10)
+    rev_reg_req3, rev_reg_reply3 = sdk_send_and_check(
+        [json.dumps(sdk_sign_request_from_dict(
+            looper, sdk_wallet_steward, rev_reg_req2['operation']))],
+        looper,
+        txnPoolNodeSet,
+        sdk_pool_handle)[0]
+    reg_delta_req = copy.deepcopy(build_get_revoc_reg_delta)
+    reg_delta_req['operation'][REVOC_REG_DEF_ID] = rev_reg_req1['operation'][REVOC_REG_DEF_ID]
+    reg_delta_req['operation'][FROM] = rev_reg_reply1['result'][TXN_TIME]
```
ashcherbakov: Do we have tests with from/to times not equal to the state update times?

indy_common/types.py
```diff
@@ -238,6 +239,37 @@ class ClientGetRevocRegDefField(MessageValidator):
...
+    )
+
+
+class ReplyRevocRegEntryValueField(MessageValidator):
+    schema = (
+        (ACCUM, NonEmptyStringField())
+    )
+
+
+class ReplyRevocRegEntryField(MessageValidator):
```
ashcherbakov: Where is it used?

indy_node/server/domain_req_handler.py
```diff
@@ -422,6 +426,122 @@ def handleGetRevocRegDefReq(self, request: Request):
...
+        # Get root hash for "to" timestamp
+        # Get REVOC_REG_ENTRY for timestamp "to"
+        path_to_reg_entry = domain.make_state_path_for_revoc_reg_entry(authors_did=author_did,
+                                                                       revoc_reg_def_id=revoc_reg_def_id)
+        past_root_to, \
+            reg_entry_to, \
+            seq_no_to, \
+            last_update_time_to, \
+            reg_entry_proof_to = self._get_reg_entry_by_timestamp(req_ts_to, path_to_reg_entry)
+        if past_root_to:
```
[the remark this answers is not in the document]
anikitinDSR: Tests for this:
 test_send_earlier_then_first_entry_by_default
test_send_earlier_then_first_entry_by_demand

indy_node/test/anon_creds/test_get_revoc_reg_delta_send_by_default.py
anikitinDSR: Tests for this:
test_send_with_from_by_default
test_send_with_from_by_demand

indy_node/server/domain_req_handler.py
```diff
@@ -422,6 +426,138 @@ def handleGetRevocRegDefReq(self, request: Request):
...
+        3. Get ISSUANCE_TYPE for REVOC_REG_DEF (revoked/issued strategy)
+        4. Compute issued and revoked indices by corresponding strategy
+        5. Make result
+           5.1 Now, if "FROM" is presented in request, then STATE_PROOF_FROM and ACCUM (revocation entry for "FROM" timestamp)
+               will added into data section
+           5.2 If not, then only STATE_PROOF for "TO" revocation entry will added
+        :param request:
+        :return: Reply
+        """
+        self._validateGetRevocRegDelta(request)
```
ashcherbakov: Why do we call it here and not in `validate`?

PR #619, opened 2018-03-21
INDY-1205: make stateTsStorage type configurable, set rocksdb by default.

sergey-shilov:
Also added rocksdb-related stuff.


hyperledger-github:

PR #623, opened 2018-03-21
INDY-1205: fix read_ledger, add rocksdb to test_ts_revocation_store.

sergey-shilov:


scripts/read_ledger
```diff
@@ -92,7 +92,7 @@ def get_ledger(type_, ledger_data_dir):
         print("Unknown ledger type: {}".format(type_))
         exit()
 
-    hash_store = LevelDbHashStore(dataDir=ledger_data_dir, fileNamePrefix=type_)
+    hash_store = initHashStore(ledger_data_dir, type_, config)
     return Ledger(CompactMerkleTree(hashStore=hash_store), dataDir=ledger_data_dir, fileName=ledger_name)
```
ashcherbakov: Read ledger for rocksdb should look differently.
It should not create a tmp copy, and should open the db in read-only mode.

sergey-shilov: Yes, I know. But I planned to make it in a separate PR as it requires changes (traversing of read_only flag through init helpers) and testing.

ashcherbakov: ok

PR #625, opened 2018-03-23
[INDY-1218] Added libsodium18 as dependency for indy-node package

anikitinDSR:

### DIFF
--- a/build-scripts/ubuntu-1604/build-indy-node.sh
+++ b/build-scripts/ubuntu-1604/build-indy-node.sh
@@ -27,6 +27,7 @@ fpm --input-type "python" \
     --exclude "*.pyo" \
     --depends at \
     --depends iptables \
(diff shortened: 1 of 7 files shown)


build-scripts/ubuntu-1604/build-indy-node.sh
```diff
@@ -27,6 +27,7 @@ fpm --input-type "python" \
     --exclude "*.pyo" \
     --depends at \
     --depends iptables \
+    --depends libsodium18 \
```
ashcherbakov: I think we can now remove libsodium18 from other places (like Vagrant) so that we test that it's installed properly.

anikitinDSR: Yes. After this fix, libsodium18 installation steps in vagrant/docker will not be needed.

PR #640, opened 2018-04-07
INDY-1226: deprecations old python cli

Toktar:
Changes:
- added message about new getting started in readme
- added message about new getting started in start of old cli script


hyperledger-github:

PR #643, opened 2018-04-10
New performance scripts

VladimirWork:
New performance scripts made by Logigear.


ashcherbakov:
test this please

scripts/performance/perf_traffic.py
```diff
@@ -0,0 +1,288 @@
...
+        Synchronize within threads to update some necessary information.
+        """
+        self.__lock.acquire()
+
+        if self.start_time == 0 and self.finish_time != 0:
+            self.start_time = self.finish_time
+
+        if self.current_total_txn != 0 and \
+                self.current_total_txn % self.transactions_delay == 0:
+            time.sleep(random.randint(1, 10))
```
ashcherbakov: 1) We need to be able to specify delay (random delay is just one of the possible options that also makes sense)
2) We need to be able to specify the type of request (nym, attr, claim_def, etc.). Random choosing of req type is just one of the options (which makes total sense).

scripts/performance/perf_traffic.py
```diff
@@ -0,0 +1,288 @@
...
+
+        :param sender: send the request.
+        :param args: contains some arguments to send request to ledger
+                     (pool handle, wallet handle, submitter did)
+        :return: response time.
+        """
+        kind = TesterSimulateTraffic._random_req_kind()
+        data = TesterSimulateTraffic._random_sample_for_get_request(kind)
+
+        req = await requests_builder.RequestBuilder.build_request(args, kind,
```
ashcherbakov: It looks like we are still waiting for the reply. So, we are waiting for the reply, then sleeping for some random time, and only then sending a new request.
We need to be able to send requests without waiting for results (sending request, say, every 1 sec).

scripts/performance/Perf_runner.py
```diff
@@ -1,118 +1,374 @@
...
+        utils.run_async_method(loop, tester.test)
+        loop.close()
+
+    def create_tester(self):
+        """
+        Create tester base mode "-a", "-t", "-g", "-l".
+
+        :return: tester
+        """
+        if self.options.adding:
```
ashcherbakov: Why do we need multiple scripts that all do almost the same? What is the difference between PerformanceTesterForAddingRequest, TesterSimulateLoad, TesterSimulateTraffic? Why can't we have just one script with parameters specifying type of requests, delay between requests, flag whether to wait for reply before sending the next request, etc.

scripts/performance/Perf_runner.py
```diff
@@ -1,118 +1,374 @@
...
+            return perf_get_requests.PerformanceTesterGetSentRequestFromLedger(
+                self.options.info_dir, self.options.kind,
+                self.options.thread_num, log=self.options.log)
+
+        elif self.options.loading:
+            return perf_load.TesterSimulateLoad(
+                self.options.clients, self.options.txns,
+                self.options.time_out, self.options.log)
+
+        elif self.options.simulate_traffic:
```
ashcherbakov: It looks like all scripts can't send requests in a real traffic mode. All of them are waiting for the result of the previous request before sending the next one.

PR #651, opened 2018-04-16
INDY-1272: added version to pip

Toktar:
changes:
-added version 9.0.3 of the library pip, because pip released version 10.0.0


hyperledger-github:

PR #661, opened 2018-04-24
[INDY-1287] fix base58 version 1.0.0

Toktar:
A new version of the base58 1.0.0 was released. The format of the response of the b58encode() and alphabet were changed from string to bytes.

Changes:

- added method decode() to calling base58.alphabet
- added method decode() to calling b58encode()


hyperledger-github:

PR #672, opened 2018-04-27
Improve ansible playbooks for test pool

skhoroshavin:

### DIFF
--- a/scripts/ansible/pool_install.yml
+++ b/scripts/ansible/pool_install.yml
@@ -1,8 +1,12 @@
 - hosts: pool
+  gather_facts: no
 
   vars:
     pool_prefix: live_node
     pool_size: "{{ groups['pool']|count }}"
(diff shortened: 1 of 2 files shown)


scripts/ansible/pool_install.yml
```diff
@@ -1,8 +1,12 @@
 - hosts: pool
+  gather_facts: no
 
   vars:
     pool_prefix: live_node
     pool_size: "{{ groups['pool']|count }}"
+    node_ver: 1.3.388
```
ashcherbakov: Maybe we should use some valid versions as default values, otherwise it's easy to forget to change them

skhoroshavin: @ashcherbakov this is version I'm currently using in test pool, so it should be valid unless I misunderstand something. What should be really done is separation of configurable variables from playbooks, which is out of scope of this PR.

PR #677, opened 2018-04-27
INDY-1289: support read-only state in read_ledger with RocksDB.

sergey-shilov:
Since read-only mode is supported for RocksDB a copy of
ledger is not created in case of RocksDB-based ledger.

Also added additional checks to avoid removing of the
ledger instead of it's copy.

### DIFF
--- a/scripts/read_ledger
+++ b/scripts/read_ledger
@@ -8,9 +8,12 @@ import logging
 import os
 import shutil
 
+from pathlib import Path
+
 from common.serializers.json_serializer import JsonSerializer
 from ledger.compact_merkle_tree import CompactMerkleTree
 from plenum.common.ledger import Ledger
+from plenum.common.constants import HS_ROCKSDB
 from storage.helper import initHashStore
 
 from indy_common.config_util import getConfig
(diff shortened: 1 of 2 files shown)


scripts/read_ledger
```diff
@@ -153,15 +156,26 @@ def make_copy_of_ledger(data_dir):
...
+    config = getConfig()
 
-    # TODO: works well only for small ledgers,
     ledger_data_dir = get_ledger_dir(args.node_name, args.client_name, args.network)
     read_copy_ledger_data_dir = None
     try:
-        read_copy_ledger_data_dir = make_copy_of_ledger(ledger_data_dir)
-        ledger = get_ledger(args.type, read_copy_ledger_data_dir)
+        # RocksDB supports real read-only mode and does not need to have a ledger copy.
+        if config.hashStore['type'].lower() != HS_ROCKSDB:
```
ashcherbakov: shouldn't it be `upper()`?

scripts/read_ledger
```diff
@@ -153,15 +156,26 @@ def make_copy_of_ledger(data_dir):
...
+            # NOTE: such approach works well only for small ledgers.
+            tmp = make_copy_of_ledger(ledger_data_dir)
```
anikitinDSR: Maybe we should check, that shutil.copytree complete successfully? What if we have really big ledger, and will raised  "There is no space left on device" error?

sergey-shilov: Hmm, I think it's already done. If exception is risen then we go directly to finally block. The read_copy_ledger_data_dir variable is empty in this case, so we do nothing, but a user will see the error message.

sergey-shilov:
Test this please.

PR #679, opened 2018-04-28
INDY-1244: set RocksDB as a key-value storage for all.

sergey-shilov:

### DIFF
--- a/indy_common/config.py
+++ b/indy_common/config.py
@@ -2,7 +2,7 @@ import os
 import logging
 from collections import OrderedDict
 
-from plenum.common.constants import ClientBootStrategy, HS_LEVELDB, KeyValueStorageType
+from plenum.common.constants import ClientBootStrategy, HS_ROCKSDB, HS_LEVELDB, KeyValueStorageType
 from plenum.config import pool_transactions_file_base, domain_transactions_file_base
 
 from indy_common.constants import Environment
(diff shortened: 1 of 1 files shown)


data/migrations/deb/1.3.396_to_1.3.397.py
```diff
@@ -0,0 +1,197 @@
...
+                     .format(rocksdb_ledger_dir, ledger_dir))
+        shutil.rmtree(rocksdb_ledger_dir)
+        return False
+
+    set_own_perm("indy", ledger_dir)
+
+    return True
+
+
+if migrate_all():
```
dsurnin: as far as I remember one of the requirements for the migration script - it should do nothing and finishes without any error if it runs several times
Also we have reinstall command that should run smoothly as well.
I think we need to determine if ledgers are leveldb or rocksdb somehow to achieve the reinstall/rerun goal.
Probably the easiest way is to add some sign to the name of the directory or something similar.

anikitinDSR:
test this please

data/migrations/deb/1.3.396_to_1.3.397.py
```diff
@@ -0,0 +1,197 @@
...
+    # Migrate transaction logs, they use integer keys
+    for ledger_type in ledger_types:
+        db_name = ledger_type + "_transactions"
+        if not migrate_storage(leveldb_ledger_dir, rocksdb_ledger_dir, db_name, True):
+            logger.error("Could not migrate {}, DB path: {}"
+                         .format(db_name, os.path.join(leveldb_ledger_dir, db_name)))
+            return False
+
+    # Migrate other storages with non-integer keys
+    for db_name in ["attr_db", "idr_cache_db", "seq_no_db", "state_signature"]:
```
ashcherbakov: What about `state_ts_db`?

anikitinDSR: This storage is already on RocksDB

PR #684, opened 2018-05-04
INDY-1173: Bugfix for command to restart pool

Toktar:
Changes:
- removed permission to restart pool by TGB
- fixed restart with cancel command
- added opportunity to send restart command with datetime="" 
- added cancel to restart now
- added tests


Toktar:
test this please

PR #688, opened 2018-05-10
[INDY-1173]Bugfix 1173 add command to restart pool

Toktar:
fix problem with datetime="" for pool_restart comman


Toktar:
test this please

PR #703, opened 2018-05-21
Recorder

devin-fisher:
Alex please begin to review the work that Lovesh has done for the recorder and replay.  

Please do not merge this pull request yet. There are few inconsistencies with the helper scripts that I need to work out tomorrow. But it is ready enough to begin reviewing. 

If there are any CI issues we would track them down tomorrow.


hyperledger-github:

tools/diagnostics/nsdiff
```diff
@@ -0,0 +1,875 @@
...
+           4. Extract the last/latest (view no, ppSeqNo) tuple from the logs
+              emitted by replica.py and node.py. TODO: update this description
+              once we have concrete examples.
+           5. All other files and directories are unaltered.
+        """
+        logger.debug("Transforming and scrubbing {} into {}".format(sin, sout))
+    
+        # The set of filenames that must exist in a directory to be assumed a
+        # LevelDB/RocksDB instance. Will be used in the following for-loop.
+        leveldb_rocksdb_filenames = [ 'CURRENT', 'LOG' ]
```
ashcherbakov: We migrated to Rocks DB now, so Leveldb is not used anymore.

tools/diagnostics/nsdiff
```diff
@@ -0,0 +1,875 @@
...
+                    logger.debug(debug_msg.format(filename))
+                    # Assume utf-8 decoding by default.
+                    # utf-8 - use standard leveldb/rocksdb iterator() and utf-8
+                    #         decoding.
+                    opened = False
+                    try:
+                        debug_msg = ("Opening {} as a LevelDB instance with"
+                                     " initKeyValueStorage...")
+                        logger.debug(debug_msg.format(filename))
+                        leveldb_rocksdb_handle = initKeyValueStorage(
```
ashcherbakov: Please note that KeyValueStorageRocksdbIntKeys is used for the ledger

tools/diagnostics/nsdiff
```diff
@@ -0,0 +1,875 @@
...
+                    # where fileNamePrefix is one of the following:
+                    #    config, pool, domain
+                    # FileHashStore requires a fileNamePrefix.
+                    fileNamePrefix = filename.split("_")[0]
+                    ledger = Ledger(
+                        CompactMerkleTree(
+                            # The FileHashStore reads
+                            # {fileNamePrefix}_merkleLeaves and
+                            # {fileNamePrefix}_merkleNodes from dataDir
+                            hashStore=FileHashStore(
```
ashcherbakov: RocksDB-based hash store is actually used by default and in production (see `initHashStore`)

tools/diagnostics/nsdiff
```diff
@@ -0,0 +1,875 @@
...
+    def _mkdir_p(self, path):
+        try:
+            os.makedirs(path)
+        except OSError as exc:  # Python >2.5
+            if exc.errno == errno.EEXIST and os.path.isdir(path):
+                pass
+            else:
+                raise
+    
+    def transform_and_scrub(self, sname, sin, sout):
```
ashcherbakov: 1) Please split the method into a number of smaller ones
2) Please avoid too may hierarchical indentation (if-else-if-else-try-if-else....)

tools/diagnostics/nsdiff
```diff
@@ -0,0 +1,875 @@
...
+            if (is_leveldb_rocksdb_dir):
+                logger.debug("{} is a LevelDB/RocksDB directory!"
+                    .format(dirName))
+
+                # Use "nodename" in place of the node's name when a directory
+                # (dirName in this case) matches the node name (sname)
+                if (sname == filename):
+                    filename = "nodename"
+
+                if (filename.endswith("_transactions")):
```
ashcherbakov: For me it looks like the method is overcomplicated. The type of the transaction log, hash store and other storages can be get from config file. Why don't we get it from there instead of trying all variants in try-else style?

tools/diagnostics/nsdiff
```diff
@@ -0,0 +1,875 @@
+#!/usr/bin/env python
+
+# README:
+#
+# This script assumes:
+# --------------------
+#
+# 1. The entirety of a node's state is bundled as a zip or tar file
```
ashcherbakov: Please specify what exactly is compared:
- txn log (for all ledgers)
- hash store (for all ledgers)
- all states
- state_signature (BLS) 
- attr_store
- seq_no_db
- ts_store
????

recorder/replay_script.py
```diff
@@ -0,0 +1,142 @@
...
+    importlib.reload(stp_zmq.kit_zstack)
+    import plenum.common.stacks
+    importlib.reload(plenum.common.stacks)
+    import plenum.server.node
+    importlib.reload(plenum.server.node)
+    import indy_node.server.node
+    importlib.reload(indy_node.server.node)
+
+
+def replay_node():
```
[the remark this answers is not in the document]
devin-fisher: It was moved. I thought I had closed this PR. See #737

tools/diagnostics/nscapture
```diff
@@ -0,0 +1,670 @@
...
+        unittest.main(argv=['capture_test'])
+        sys.exit(0)
+
+    def _touch(path, content):
+        d = os.path.dirname(path)
+        os.makedirs(d, exist_ok=True)
+        with open(path, 'w') as f:
+            f.write(content)
+
+    class TestCapture(unittest.TestCase):
```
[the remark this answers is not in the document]
devin-fisher: nscapture was written as a stand-alone script. It is not a python module. It is outside the view of pytest. I choose unittest so there was no dependency to run the test. Also, unittest was easier to run via a command line option (-t). I'm welcome to discuss how to do this better but simply putting them in a directory and using pytest would not accomplish my portable stand-alone CLI script requirement requirements.

The tests are exercised in the CI system in indy_node/test/tools/test_nsrepay.py

tools/diagnostics/nsreplay
```diff
@@ -0,0 +1,416 @@
...
+        logger.debug("Temporary replay node directory {}".format(
+            replay_node_dir))
+        # with tempfile.TemporaryDirectory() as replay_node_dir:
+        general_config_dir = create_config_dirs(replay_node_dir)
+        config = getConfig(general_config_dir)
+        self.update_loaded_config(config)
+        pool_dir = join(replay_node_dir, pool_name)
+        orig_node_pool_dir = self.s1dir
+
+        trg_var_dir = join(replay_node_dir, 'var', 'lib', 'indy', pool_name)
```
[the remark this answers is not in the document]
devin-fisher: @ckochenower Yeah let definitely use it. I'm using it in nscapture.

tools/diagnostics/nsdiff
```diff
@@ -0,0 +1,875 @@
...
+# Installation:
+# -------------
+#
+# If this script is used on a server or workstation that has indy-node installed
+# and indy-node is known to be working, you do not need to install anything.
+# Otherwise, the following information will be helpful.
+#
+# This script depends on leveldb/rocksdb python bindings which in turn rely on
+# the leveldb C runtime libs and headers. You must install leveldb/rocksdb
+# dependencies. See Installation > Prerequisites section below.
```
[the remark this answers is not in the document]
devin-fisher: These tools are not intended as solely as development tools. They are stand-alone tools. Their requirement should be equal or less than to running the node. 

So they may be duplication since they would be installed with the node.

tools/diagnostics/nsdiff
devin-fisher: The script was written during the transition. Do you think it worth ripping Leveldb? Have we ripped out LevelDB in the node code? Is it no longer possible to run the node with LevelDB? If so, we should rip it out. @ckochenower

tools/diagnostics/nsdiff
devin-fisher: We can put some TODO comment to do this. But I don't think we should hold up the PR till this is done.
Alex, I agree with the suggestions but are you ok if we put it on our TODO list?

tools/diagnostics/nsdiff
devin-fisher: Initial we were not capturing the configuration. (not part of nsdiff original scope) So that is why we had a try-fail cycle. I think the try-fail gives us better resiliency at not much cost in the complexity of the code. But we can add TODOs to simplify this. But I don't think that this should hold this back.

PR #715, opened 2018-05-23
[WIP][skip ci][INDY-1363] add unsigned and signed action to ActionReqHandler

Toktar:
Previously, the sending of the validator info command could only be signed. Now this feature is eliminated.

Changes:
- added unsigned_action_types and signed_action_types to ActionReqHandler


Toktar:
will not be implemented

PR #730, opened 2018-05-29
State proof optmz

lovesh:


indy_node/server/domain_req_handler.py
```diff
@@ -421,18 +421,18 @@ def add_query_handler(self, txn_type, handler: Callable):
...
+            nym_data = self.stateSerializer.deserialize(nym_data)
+            nym_data[TARGET_NYM] = nym
+            data = self.stateSerializer.serialize(nym_data)
+            seq_no = nym_data[f.SEQ_NO.nm]
+            update_time = nym_data[TXN_TIME]
         else:
             data = None
             seq_no = None
-            proof = self.make_proof(path)
+            proof = None
```
ashcherbakov: We need to have proof of non-existence here

indy_node/server/domain_req_handler.py
```diff
@@ -421,18 +421,18 @@ def add_query_handler(self, txn_type, handler: Callable):
 
     def handleGetNymReq(self, request: Request):
         nym = request.operation[TARGET_NYM]
-        nymData = self.idrCache.getNym(nym, isCommitted=True)
```
ashcherbakov: Why don't we use IDR Cache anymore?

lovesh: Because same data is present in state trie too.

indy_node/server/domain_req_handler.py
lovesh: Adding

lovesh:
@ashcherbakov This PR has DCO issues and bad commits so closing, created https://github.com/hyperledger/indy-node/pull/735 with suggested changes

PR #731, opened 2018-05-29
fix in migration script

ashcherbakov:


data/migrations/deb/1_3_428_to_1_3_429.py
```diff
@@ -78,7 +78,7 @@ def migrate_txn_log(db_dir, db_name):
     try:
         for key, val in src_storage.iterator():
             val = ledger_txn_serializer.deserialize(val)
-            new_val = transform_to_new_format(txn=val, seq_no=key)
+            new_val = transform_to_new_format(txn=val, seq_no=int(key))
```
anikitinDSR: Do we need to decode before casting into int?
like `int(key.decode())`

PR #735, opened 2018-05-31
refactor state proof generation

lovesh:
Signed-off-by: Lovesh Harchandani <lovesh.bond@gmail.com>

### DIFF
--- a/indy_node/server/domain_req_handler.py
+++ b/indy_node/server/domain_req_handler.py
@@ -285,7 +285,7 @@ class DomainReqHandler(PHandler):
         cred_def_path = domain.make_state_path_for_claim_def(authors_did=tags[0],
                                                              signature_type=tags[2],
                                                              schema_seq_no=tags[3])
-        cred_def, _, _, _ = self.lookup(cred_def_path, isCommitted=False, get_proof=False)
+        cred_def, _, _, _ = self.lookup(cred_def_path, isCommitted=False)
         if cred_def is None:
             raise InvalidClientRequest(req.identifier,
                                        req.reqId,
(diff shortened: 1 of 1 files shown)


indy_node/server/domain_req_handler.py
```diff
@@ -423,18 +423,17 @@ def add_query_handler(self, txn_type, handler: Callable):
...
         nym = request.operation[TARGET_NYM]
-        nymData = self.idrCache.getNym(nym, isCommitted=True)
         path = domain.make_state_path_for_nym(nym)
-        if nymData:
-            nymData[TARGET_NYM] = nym
-            data = self.stateSerializer.serialize(nymData)
-            seq_no = nymData[f.SEQ_NO.nm]
-            update_time = nymData[TXN_TIME]
-            proof = self.make_proof(path)
+        nym_data, proof = self.get_value_from_state(path, with_proof=True)
```
ashcherbakov: So, can we remove `idrCache` at all then? Why do we need it?

indy_node/server/domain_req_handler.py
```diff
@@ -423,18 +423,17 @@ def add_query_handler(self, txn_type, handler: Callable):
...
+        if nym_data:
+            nym_data = self.stateSerializer.deserialize(nym_data)
+            nym_data[TARGET_NYM] = nym
+            data = self.stateSerializer.serialize(nym_data)
+            seq_no = nym_data[f.SEQ_NO.nm]
+            update_time = nym_data[TXN_TIME]
         else:
```
ashcherbakov: Please generate the proof on non-existence here

indy_node/server/domain_req_handler.py
lovesh: For validation, we dont query trie since its expensive to query

indy_node/server/domain_req_handler.py
lovesh: Its there in `nym_data, proof = self.get_value_from_state(path, with_proof=True)`. I dont see an existing test for proof of absence, i can add one

lovesh: Added

PR #737, opened 2018-05-31
Node scripts

devin-fisher:
Replacement for the recorder PR. This has the correct files and fixes DCO issues


hyperledger-github:

devin-fisher:
This will not pass CI until the recorder PR in plenum is merged

tools/diagnostics/nscapture
```diff
@@ -0,0 +1,782 @@
...
+        raise
+
+    try:
+        do_capture(args, files)
+    except:
+        logger.error('Unable to build capture archive')
+        raise
+
+
+# ***************
```
devin-fisher: From Alex:
We use pytest for testing. Can we switch to pytest to execute the test during CI?

devin-fisher: And please move tests to a separate folder (like in other places in code)

devin-fisher: nscapture was written as a stand-alone script. It is not a python module. It is outside the view of pytest. I choose unittest so there was no dependency to run the test. Also, unittest was easier to run via a command line option (-t). I'm welcome to discuss how to do this better but simply putting them in a directory and using pytest would not accomplish my portable stand-alone CLI script requirement requirements.

The tests are exercised in the CI system in indy_node/test/tools/test_nsrepay.py

tools/diagnostics/nsdiff
```diff
@@ -0,0 +1,875 @@
...
+#
+# 1. The entirety of a node's state is bundled as a zip or tar file
+# 2. The zip/tar file has the name <node_name>.zip or <node_name>.tar.gz where
+#    <node_name> is the name of the node.
+# 3. Both zip/tar files to be compared are on the filesystem where nsdiff is
+#    installed. In other words, nsdiff does not attempt to connect or collect
+#    state data from remote nodes.
+# 4. nscapture is/was used to capture state passed as state1 and state2
+#
+# Installation:
```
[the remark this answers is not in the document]
devin-fisher: These tools are not intended as solely as development tools. They are stand-alone tools. Their requirement should be equal or less than to running the node.

So they may be duplication since they would be installed with the node.

tools/diagnostics/nsdiff
```diff
@@ -0,0 +1,875 @@
...
+# python3 or pip3 imply Python 3.5.2.
+# 
+# Note that anytime pip/pip3 is used to install python packages, the
+# python/python3 environment is changed without the knowledge of package
+# managers (apt, yum, brew, etc.). Side-effects of pip installations and
+# os-controlled updates (i.e. automatic updates) to your system's python have
+# the potential to cause instability/breakage of either your system or
+# development environment. Most modern operating systems depend on Python. If
+# you would prefer not to make changes to your system python/python3, please
+# consider setting up a Python Virtual Environment (highly recommended):
```
devin-fisher: FROM ALEX
I think in most of the cases this script will be run on a real installation from deb packages (for example in Docker), not with local nodes.

tools/diagnostics/nsdiff
```diff
@@ -0,0 +1,875 @@
...
+           4. Extract the last/latest (view no, ppSeqNo) tuple from the logs
+              emitted by replica.py and node.py. TODO: update this description
+              once we have concrete examples.
+           5. All other files and directories are unaltered.
+        """
+        logger.debug("Transforming and scrubbing {} into {}".format(sin, sout))
+    
+        # The set of filenames that must exist in a directory to be assumed a
+        # LevelDB/RocksDB instance. Will be used in the following for-loop.
+        leveldb_rocksdb_filenames = [ 'CURRENT', 'LOG' ]
```
devin-fisher: FROM ALEX
We migrated to Rocks DB now, so Leveldb is not used anymore.

devin-fisher: The script was written during the transition. Do you think it worth ripping Leveldb? Have we ripped out LevelDB in the node code? Is it no longer possible to run the node with LevelDB? If so, we should rip it out. @ckochenower

tools/diagnostics/nsdiff
```diff
@@ -0,0 +1,875 @@
...
+                else:
+                    debug_msg = ("{} is NOT a transaction LevelDB/RocksDB"
+                                 "instance!")
+                    logger.debug(debug_msg.format(filename))
+                    # Assume utf-8 decoding by default.
+                    # utf-8 - use standard leveldb/rocksdb iterator() and utf-8
+                    #         decoding.
+                    opened = False
+                    try:
+                        debug_msg = ("Opening {} as a LevelDB instance with"
```
devin-fisher: FROM ALEX:
Please note that KeyValueStorageRocksdbIntKeys is used for the ledger

tools/diagnostics/nsdiff
```diff
@@ -0,0 +1,875 @@
...
+                    # where fileNamePrefix is one of the following:
+                    #    config, pool, domain
+                    # FileHashStore requires a fileNamePrefix.
+                    fileNamePrefix = filename.split("_")[0]
+                    ledger = Ledger(
+                        CompactMerkleTree(
+                            # The FileHashStore reads
+                            # {fileNamePrefix}_merkleLeaves and
+                            # {fileNamePrefix}_merkleNodes from dataDir
+                            hashStore=FileHashStore(
```
devin-fisher: FROM ALEX:
RocksDB-based hash store is actually used by default and in production (see initHashStore)

tools/diagnostics/nsdiff
```diff
@@ -0,0 +1,875 @@
...
+    def _mkdir_p(self, path):
+        try:
+            os.makedirs(path)
+        except OSError as exc:  # Python >2.5
+            if exc.errno == errno.EEXIST and os.path.isdir(path):
+                pass
+            else:
+                raise
+    
+    def transform_and_scrub(self, sname, sin, sout):
```
devin-fisher: FROM ALEX
- Please split the method into a number of smaller ones
- Please avoid too may hierarchical indentation (if-else-if-else-try-if-else....)

devin-fisher: We can put some TODO comment to do this. But I don't think we should hold up the PR till this is done.
Alex, I agree with the suggestions but are you ok if we put it on our TODO list?

tools/diagnostics/nsdiff
```diff
@@ -0,0 +1,875 @@
...
+                subdirList[:] = []
+                continue
+
+            if (is_leveldb_rocksdb_dir):
+                logger.debug("{} is a LevelDB/RocksDB directory!"
+                    .format(dirName))
+
+                # Use "nodename" in place of the node's name when a directory
+                # (dirName in this case) matches the node name (sname)
+                if (sname == filename):
```
devin-fisher: FROM ALEX:
For me it looks like the method is overcomplicated. The type of the transaction log, hash store and other storages can be get from config file. Why don't we get it from there instead of trying all variants in try-else style?

devin-fisher: Initial we were not capturing the configuration. (not part of nsdiff original scope) So that is why we had a try-fail cycle. I think the try-fail gives us better resiliency at not much cost in the complexity of the code. But we can add TODOs to simplify this. But I don't think that this should hold this back.

tools/diagnostics/nsdiff
```diff
@@ -0,0 +1,875 @@
+#!/usr/bin/env python
+
+# README:
+#
+# This script assumes:
```
devin-fisher: FROM ALEX:
Please specify what exactly is compared:

- txn log (for all ledgers)
- hash store (for all ledgers)
- all states
- state_signature (BLS)
- attr_store
- seq_no_db
- ts_store
- ????

devin-fisher: @ckochenower added some doc in the comments to address this.

tools/diagnostics/nsdiff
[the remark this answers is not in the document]
ashcherbakov: But you can reference to the documentation in indy-node instead of duplicating it. We should support it up-to-date in one place only.

tools/diagnostics/nscapture
ashcherbakov: OK. We can create a ticket for more enhancements.

tools/diagnostics/nsdiff
ashcherbakov: It's still possible to run the code with leveldb. The DB is defined in config file.

tools/diagnostics/nsdiff
ashcherbakov: Well, if this takes too much time for re-factroing then fine, we can do it later with a ticket and TODO comment. However, I would prefer to have a cleaner and more readable code (maybe do it together with the next item)?

ashcherbakov:
test this please

devin-fisher:
test this please

devin-fisher:
(ci) test this please

ashcherbakov:
(ci) test this please

ashcherbakov:
(ci) test this please

PR #748, opened 2018-06-06
[INDY-1370] Сhange key in requests map

Toktar:
Request field "key" was changed from (identifier, reqId) to digest, therefore use of key has been changed.
In addition, map "requestSender" key  was changed too and its use has been update.


Toktar:
test this please

PR #755, opened 2018-06-09
INDY-1372: add optional 'tag' field to CLAIM_DEF

ashcherbakov:
- added optional `tag` field to CLAIM_DEF
- use utility methods for SCHEMA and CLAIM_DEF 
- added more tests


indy_client/anon_creds/indy_public_repo.py
```diff
@@ -38,7 +40,7 @@ def _ensureReqCompleted(reqKey, client, clbk):
 
 
 def _getData(result, error):
-    data = result.get(DATA) if result.get(DATA) else {}
+    data = result.get(DATA) or {}
```
dsurnin: You can provide default value as a second arg to get
result.get(DATA, {})

PR #768, opened 2018-06-15
Update per_processes.py to handle perfomance testing needs

dsurnin:

### DIFF
--- a/scripts/performance/perf_processes.py
+++ b/scripts/performance/perf_processes.py
@@ -1,3 +1,5 @@
+from abc import ABCMeta, abstractmethod
+import copy
 import shutil
 import json
 import time
@@ -14,7 +16,7 @@ import base58
 import libnacl
 import random
 
-from indy import pool, wallet, did, ledger, anoncreds
+from indy import pool, wallet, did, ledger, anoncreds, blob_storage
 
 
 parser = argparse.ArgumentParser(description='The script generates bunch of txns for the pool with Indy SDK.')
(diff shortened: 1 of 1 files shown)


scripts/performance/perf_processes.py
```diff
@@ -169,12 +178,297 @@ def dump_stat(self, dump_all: bool = False):
...
+    def _rand_data(self):
+        return self.random_string(32)
+
+    def _from_file_str_data(self, file_str):
+        return file_str
+
+    def _gen_req_data(self):
+        if self.file_name is not None:
+            with open(self.file_name, "rt") as data_file:
+                while True:
```
ashcherbakov: Where do we stop the loop? Is it intended that there is `while True` here?

scripts/performance/perf_processes.py
```diff
@@ -56,10 +58,9 @@ def check_seed(seed: str):
                     type=check_seed, required=False, dest='seed')
 
 parser.add_argument('-k', '--kind',
```
ashcherbakov: Please provide documentation with real example on how to use the script (consider multiple examples and modes)

scripts/performance/perf_processes.py
```diff
@@ -56,10 +58,9 @@ def check_seed(seed: str):
...
-                    help='Kind of request kind to use. One of ["nym", "schema", "attrib", "rand"]'
-                         'Default value is "nym"',
-                    default="nym", choices=['nym', 'schema', 'attrib', 'rand'],
-                    required=False, dest='req_kind')
+                    help='Kind of request to send. One of ["nym", "schema", "attrib", "get_nym"] '
+                         'Default value is "nym". Could be combined in form of "{req_type1: num, ...}" ',
```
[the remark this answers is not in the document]
dsurnin: it will be a huge description and I think format could be changed after QA will check it in details

PR #779, opened 2018-06-21
Public/fix get get revoc and migration

anikitinDSR:


indy_node/test/anon_creds/test_get_revoc_reg_entry.py
```diff
@@ -1,6 +1,30 @@
...
+def test_get_revoc_reg_entry_without_any_entry_return_none(send_revoc_reg_def_by_default,
+                                                           build_get_revoc_reg_entry,
+                                                           txnPoolNodeSet):
+    req = build_get_revoc_reg_entry
+    req_handler = txnPoolNodeSet[0].getDomainReqHandler()
+    result = req_handler.handleGetRevocRegReq(SafeRequest(**req))
+    assert result[DATA] is None
+    assert result[f.SEQ_NO.nm] is None
+    assert result[TXN_TIME] is None
+    assert STATE_PROOF not in result
```
ashcherbakov: STATE_PROOF must be present even when there is no value returned (proof of non-existence).

PR #806, opened 2018-07-04
Add field settings for file reader; Fix cred_def and REG_ENTRY_REVOC

dsurnin:


docs/process-based-load-script.md
```diff
@@ -8,10 +8,10 @@
 ## Requirements
 
 * native libs
-    * libindy >= 1.4.0~588 (master repo)
+    * libindy >= 1.4.0~626 (master repo)
 
 * python lib
-    * python3-indy >= 1.4.0.dev588
+    * python3-indy >= 1.5.0.dev626
```
ashcherbakov: Should it be 1.4.0?

dsurnin: script uses master branch, it is 1.5.0

scripts/performance/perf_processes.py
```diff
@@ -302,8 +317,13 @@ def _rand_data(self):
 
     def _from_file_str_data(self, file_str):
         req_json = super()._from_file_str_data(file_str)
-        req_did = json.loads(req_json)['result']['txn']['data']['dest']
-        return req_did
+        if req_json is None:
+            return None
+        tmp = json.loads(req_json)
+        txn = tmp.get('result', {}).get('txn', {}) or tmp.get('txn', {})
```
ashcherbakov: This looks like duplicate code in multiple places. Can we use some utility/helper method?

scripts/performance/perf_processes.py
```diff
@@ -323,7 +343,13 @@ class RGSchema(RequestGenerator):
...
     def _from_file_str_data(self, file_str):
         req_json = super()._from_file_str_data(file_str)
-        schema_id = json.loads(req_json)['result']['txnMetadata']['txnId']
+        if req_json is None:
+            return None
+        tmp = json.loads(req_json)
+        txn_type = (tmp.get('result', {}).get('txn', {}) or tmp.get('txn', {})).get('type', None)
+        if txn_type not in ["101", "107"]:
+            return None
+        schema_id = (tmp.get('result', {}).get('txnMetadata', {}) or tmp.get('txnMetadata', {})).get('txnId', None)
```
ashcherbakov: This looks like duplicate code in multiple places. Can we use some utility/helper method?

scripts/performance/perf_processes.py
```diff
@@ -470,42 +521,74 @@ def __init__(self, *args, **kwargs):
             "sex": {"raw": "value1", "encoded": "123"},
             "height": {"raw": "value1", "encoded": "456"}
         })
+        self._max_cred_num = None
+        self._submitter_did = None
+        self._pool_handle = None
+        self._cred_offer_json = None
+        self._tails_writer = None
+
+    async def _upd_revoc_reg(self):
```
ashcherbakov: So, GetRevocRegEntry request consists of GET and WRITE every time, right?
Should we mention this specific in docs, so that QA can calculate the expected load from the script appropriately?

scripts/performance/perf_processes.py
```diff
@@ -751,6 +842,8 @@ def req_send(self, start_new_batch: bool = False):
         self._closing = True
         if len(self._send_q) > 0:
             await asyncio.gather(*self._send_q, return_exceptions=True)
+        if len(self._gen_q) > 0:
```
ashcherbakov: Why do we need this line now and didn't need it before?

dsurnin: this line was added to prevent crashing during the finalization step in case of time consuming requests like revoc_reg

PR #807, opened 2018-07-04
Deprecated Vagrant method for running test network

ryanwest6:
To install and run a test network of nodes on one's computer for development purposes, there are several methods available to use. However, lately only Docker has been supported. Vagrant is more difficult to use, and in many cases breaks completely. Several new, interested developers have been caught trying to use Vagrant and failing because they should be following the more-updated Docker instructions instead.

To fix this problem, Vagrant needs to be marked as deprecated in various places in indy-node's Github repository.

Signed-off-by: ryanwest6 <31378238+ryanwest6@users.noreply.github.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,8 +104,6 @@ to understand what needs to be done to create a Network, initialize and start No
 
 The described process is automated in one of the ways below (it allow to install a test Network):
 
- - **Automated VM Creation with Vagrant** [Create virtual machines](environment/vagrant/training/vb-multi-vm/TestIndyClusterSetup.md) using VirtualBox and Vagrant.
-
  - **Docker** [Start Pool and Client with Docker](environment/docker/pool/README.md)
 
  - **Docker-based pool using with new libindy-based CLI**:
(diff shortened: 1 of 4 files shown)


hyperledger-github:

hyperledger-github:

ashcherbakov:
Are we sure about a requirement to deprecate Vagrant? Maybe we should rather fix the scripts?
I'm also not sure that `CloudFormation` works properly.

ryanwest6:
Nathan, I, and others at Sovrin have discussed that it makes more sense to offer one widely support and documented way of running the test network. None of the devs at Sovrin have successfully been able to run it, but Docker works well for all of us. 

I removed it from the newbie documentation and labeled it as 'deprecated' elsewhere, but it still exists for those who want to use it. (And if someone wants to fix the scripts and get it working again, by all means they can go ahead).

ryanwest6:
None of us have used `CloudFormation` either so it may also be worth deprecating.

ashcherbakov:
(ci) test this please

ryanwest6:
Looks like the continuous-integration step is failing now, but clicking on `Details` doesn't load.

ashcherbakov:
(ci) test this please

PR #810, opened 2018-07-06
[INDY-1443] fix --json behaviour

anikitinDSR:
Changes:

-  For now --json option returns completely parsable JSON.
- Also, --json options returns all acceptable info (highest verbosity).
- Fix errors if node_info files are empty.

### DIFF
--- a/scripts/validator-info
+++ b/scripts/validator-info
@@ -570,9 +570,7 @@ def nagios(vstats):
     return "\n".join(lines);
 
 
-def get_stats_from_file(fpath, verbose, _json, _nagios):
-    with open(fpath) as f:
-        stats = json.loads(f.read())
+def get_stats_from_file(stats, verbose, _json, _nagios):
 
     logger.debug("Data {}".format(stats))
     vstats = ValidatorStats(stats, verbose)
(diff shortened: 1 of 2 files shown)


scripts/validator-info
```diff
@@ -794,19 +795,47 @@ def main():
     if not info_paths:
         print('There are no info files in {}'.format(args.basedir))
         return
-    for file_path in info_paths:
-        logger.info("Reading file {} ...".format(file_path))
-        if args.verbose:
+    if args.json:
```
ashcherbakov: So, do we always print additional data in json mode? What if this is not verbose mode?

scripts/validator-info
```diff
@@ -794,19 +795,47 @@ def main():
...
+                except json.JSONDecodeError:
+                    print("File {} has an invalid json format".format(file_path))
+                    continue
+                output_data.update(json_data)
+        if output_data:
+            print(json.dumps(output_data))
+        sys.exit(0)
+
+    for file_path in info_paths:
+        with open(file_path) as f_stats:
```
ashcherbakov: Looks like duplicate code as above

scripts/validator-info
anikitinDSR: according with Corin's comment, --json options should be the highest verbosity.
https://jira.hyperledger.org/browse/INDY-1443?focusedCommentId=46831&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-46831

PR #819, opened 2018-07-13
Experiments with network stack params

dsurnin:
Experiments with network stack params

### DIFF
--- a/indy_common/config.py
+++ b/indy_common/config.py
@@ -102,3 +102,7 @@ agentLoggingLevel = logging.INFO
 default logging level for node
 '''
 logLevel = logging.INFO
+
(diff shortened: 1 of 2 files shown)


indy_common/config.py
```diff
@@ -102,3 +102,6 @@
 default logging level for node
 '''
 logLevel = logging.INFO
+
+ZMQ_CLIENT_QUEUE_SIZE = 3000  # messages (0 - no limit)
```
ashcherbakov: Why do we set it here and not in plenum?

dsurnin: Just for convenience - only one build instead of two: plenum, node with plenum

PR #821, opened 2018-07-16
INDY-1199: Perform restart when inconsistent 3PC state is suspected

skhoroshavin:


indy_node/server/node.py
```diff
@@ -106,6 +106,9 @@ def getPoolConfig(self):
     def initPoolManager(self, ha, cliname, cliha):
         HasPoolManager.__init__(self, ha, cliname, cliha)
 
+    def on_inconsistent_3pc_state(self):
```
ashcherbakov: Please add a test that restart actually happens when we notice inconsistent 3PC state and that the node joins the pool successfully after this.

PR #828, opened 2018-07-18
Node State tools/scripts require python 3

ckochenower:
Signed-off-by: Corin Kochenower <corin.kochenower@evernym.com>


hyperledger-github:

PR #833, opened 2018-07-19
Fix some grammar and wording

PanCakeConnaisseur:
This should make the README.md easier to read/understand. Only went through roughly the first half of the document.


hyperledger-github:

PR #860, opened 2018-08-02
INDY-1528: Additional changes

ArtObr:


indy_common/auth.py
```diff
@@ -67,7 +67,8 @@ def generate_auth_map(valid_roles, need_permission=None):
         '{}_<any>_<any>_<any>'.format(VALIDATOR_INFO):
             {TRUSTEE: [], STEWARD: []},
     }
-    if need_permission is False or (need_permission is None and not getConfig().WRITES_REQUIRE_TRUST_ANCHOR):
+    if anyone_can_write is True or (anyone_can_write is None and getConfig().ANYONE_CAN_WRITE):
+        auth_map['{}_role__'.format(NYM)][None] = []
```
ashcherbakov: This rule is only about creation of new NYMs with empty role (common users). What about other roles? We need to clarify the requirement.

indy_node/test/anon_creds/test_send_write_permission_anyone.py
```diff
@@ -0,0 +1,88 @@
...
+    sdk_get_and_check_replies, sdk_sign_and_submit_req
+from plenum.common.exceptions import RequestRejectedException
+
+from indy_common.auth import Authoriser
+from indy.anoncreds import issuer_create_schema
+from indy.ledger import build_schema_request
+
+from indy_node.test.anon_creds.conftest import claim_def
+from indy_client.test.test_nym_attrib import attributeData, attributeName, attributeValue
+
```
ashcherbakov: Why is it in `anoncreds` module?

indy_node/test/anon_creds/test_send_write_permission_anyone.py
```diff
@@ -0,0 +1,88 @@
...
+                             txnPoolNodeSet,
+                             sdk_wallet_client,
+                             sdk_wallet_trust_anchor,
+                             sdk_pool_handle,
+                             attributeData):
+    # client can create another client NYM when ANYONE_CAN_WRITE set to True
+    sdk_new_client_wallet = sdk_add_new_nym(looper, sdk_pool_handle, sdk_wallet_client)
+    _, new_client_did = sdk_new_client_wallet
+
+    # client as an owner can add attribute to his NYM
```
ashcherbakov: It's better to move the checks for ATTRIBs into a separate test method. Keep tests as small as possible.

indy_node/test/anon_creds/test_send_write_permission_anyone.py
```diff
@@ -0,0 +1,88 @@
...
+    request = looper.loop.run_until_complete(build_schema_request(identifier, schema_json))
+    sdk_get_and_check_replies(looper, [sdk_sign_and_submit_req(sdk_pool_handle, sdk_wallet_trust_anchor, request)])
+
+    # Client can create schema if ANYONE_CAN_WRITE flag set to False
+    _, identifier = sdk_wallet_client
+    _, schema_json = looper.loop.run_until_complete(
+        issuer_create_schema(identifier, "name", "1.0", json.dumps(["first", "last"])))
+    request = looper.loop.run_until_complete(build_schema_request(identifier, schema_json))
+    sdk_get_and_check_replies(looper, [sdk_sign_and_submit_req(sdk_pool_handle, sdk_wallet_client, request)])
+
```
ashcherbakov: What about revocation txns? Can client send them?

indy_common/auth.py
ashcherbakov: If any NYMs, SCHEMAs, CLAIM_DEFs, and revocation txns can be created with ANYONE_CAN_WRITE=True, then maybe we should not modify this map, but rather add a check to `authorised`, so that if `typ` is one from the list above, then return True immediately.

indy_node/test/anon_creds/test_send_write_permission_anyone.py
ArtObr: Ok

indy_node/test/anon_creds/test_send_write_permission_anyone.py
ArtObr: He can. Relevant tests will be added

indy_common/auth.py
ArtObr: asked

ArtObr:
test this please

ashcherbakov:
test this please

ashcherbakov:
test this please

ArtObr:
test this please

PR #868, opened 2018-08-03
Add 2 steward utility files

mgbailey:
current_validators gets the pool ledger from read_ledger, and
parses them to condense the information to the latest valid
entries for each field. Output is to console, as json or as csv.

node_address_list will take the json output from
current_validators and distill it down to just the currently
active validators, and returns to console just the alias and the
node IP address for each, useful for making firewall whitelists.

Signed-off-by: Mike Bailey <mike.bailey@evernym.com>

### DIFF
--- a/scripts/current_validators
+++ b/scripts/current_validators
@@ -1,0 +1,138 @@
+#!/usr/bin/python3
+
+import string
+import json
+import csv
+import logging
+import argparse
+import subprocess
+import sys
+import collections
+
+logging.basicConfig(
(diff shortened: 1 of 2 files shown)


scripts/current_validators
```diff
@@ -0,0 +1,138 @@
+#!/usr/bin/python3
+
```
ashcherbakov: If the scripts need to be installed with the node package, they should be added to `setup.py` (`scripts` field).

PR #869, opened 2018-08-06
Extend upgrade command

dsurnin:


indy_node/server/config_req_handler.py
```diff
@@ -55,6 +57,20 @@ def _doStaticValidationPoolUpgrade(self, identifier, reqId, operation):
...
+        if pkg_name == "indy-node":
+            return Upgrader.getVersion(), ["indy-node"]
+        return NodeControlUtil.curr_pkt_info(pkg_name)
+
+    def get_dependencies(self, pkg_name, version):
+        base_deps = ["indy-node", "indy-plenum"]
+        if pkg_name == "indy-node":
+            return base_deps
+        deps = []
+        NodeControlUtil.dep_tree_traverse(
```
ashcherbakov: Are we going to call 3d party tools to get dependencies during the static validation? I'm not sure that this is a good option...

indy_node/server/config_req_handler.py
```diff
@@ -70,7 +86,23 @@ def validate(self, req: Request):
                 req.reqId,
                 "Nym {} not added to the ledger yet".format(origin))
         if typ == POOL_UPGRADE:
-            currentVersion = Upgrader.getVersion()
+            pkt_to_upgrade = req.operation.get(PACKAGE, getConfig().UPGRADE_ENTRY)
+            if pkt_to_upgrade:
+                currentVersion, cur_deps = self.curr_pkt_info(pkt_to_upgrade)
```
ashcherbakov: `curr_pkt_info` returns a version only, doesn't it?

indy_node/test/upgrade/test_upgrade_top_level_def_cfg.py
```diff
@@ -9,7 +9,6 @@ def transform(tool):
 
     nct = NCT(backup_dir=tdir, backup_target=tdir, transform=transform)
     try:
-        assert nct.tool.ext_upgrade is False
```
ashcherbakov: Why is it removed?

indy_node/test/upgrade/test_upgrade_top_level_ext_pkg.py
```diff
@@ -32,7 +32,6 @@ def transform(tool):
     try:
         assert EXT_PKT_VERSION, EXT_PKT_DEPS == nct.tool._ext_info()
         nct.tool._ext_init()
-        assert nct.tool.ext_upgrade is True
```
ashcherbakov: Why is it removed?

indy_node/utils/node_control_tool.py
```diff
@@ -45,7 +42,7 @@ def __init__(
         self.config = config or getConfig()
 
         assert self.config.UPGRADE_ENTRY, "UPGRADE_ENTRY config parameter must be set"
-        self.ext_upgrade = self.config.UPGRADE_ENTRY not in PACKAGES_TO_HOLD
+        self.upgrade_entry = self.config.UPGRADE_ENTRY
```
ashcherbakov: Why is it always from config? Why not from POOL_UPGRADE command?

indy_node/server/config_req_handler.py
```diff
@@ -55,6 +57,20 @@ def _doStaticValidationPoolUpgrade(self, identifier, reqId, operation):
 
         # TODO: Check if cancel is submitted before start
 
+    def curr_pkt_info(self, pkg_name):
+        if pkg_name == "indy-node":
```
ashcherbakov: Do we have a test when package name is not equal to indy-node? Can we have all existing tests parametrized so that we test both indy-node and a custom package?

indy_node/server/config_req_handler.py
```diff
@@ -55,6 +57,20 @@ def _doStaticValidationPoolUpgrade(self, identifier, reqId, operation):
...
+            return Upgrader.getVersion(), ["indy-node"]
```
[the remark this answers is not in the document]
dsurnin: only indy-node version is handled by default. to get top level package version we need to ask packet manager

indy_node/server/config_req_handler.py
dsurnin: yes we are for now. if we do not check it here will need to check it during the upgrade and raise ex, but we just told that txn was validated and fine... I think it could be strange for user...
But I don't mind to remove it. Should I?

indy_node/server/config_req_handler.py
dsurnin: version and list of dependencies

indy_node/test/upgrade/test_upgrade_top_level_def_cfg.py
dsurnin: flag was removed from the class at all

indy_node/test/upgrade/test_upgrade_top_level_ext_pkg.py
dsurnin: flag was removed from the class at all

indy_node/utils/node_control_tool.py
dsurnin: this is initialization stage. proper name will be provided by the node via upgrade message

indy_node/server/config_req_handler.py
dsurnin: we have some

indy_node/test/upgrade/test_forced_upgrade_if_request_received_after_propagate.py
```diff
@@ -40,11 +40,13 @@ def patched_process_request(request: Request, frm: str):
...
+    init_len = len(list(slow_node.upgrader._actionLog))
+
     sdk_ensure_upgrade_sent(looper, sdk_pool_handle, sdk_wallet_trustee,
                             validUpgradeExpForceTrue)
 
     looper.runFor(waits.expectedUpgradeScheduled())
 
     checkUpgradeScheduled([slow_node], validUpgradeExpForceTrue[VERSION])
-    assert len(list(slow_node.upgrader._actionLog)) == 1
+    assert len(list(slow_node.upgrader._actionLog)) > init_len
```
ashcherbakov: Can we check that it equals to `init_len  + 1`?

indy_node/test/upgrade/conftest.py
```diff
@@ -16,46 +19,89 @@ def nodeIds(nodeSet):
...
         schedule[i] = datetime.isoformat(startAt)
         startAt = startAt + timedelta(seconds=acceptableDiff + 3)
-    return dict(name='upgrade-13', version=bumpedVersion(), action=START,
-                schedule=schedule,
-                # sha256=get_valid_code_hash(),
+
+    vers = None
+
+    if request.param != APP_NAME:
+        EXT_PKT_VERSION = '7.88.999'
```
ashcherbakov: It looks like it can be moved to a helper method, as there are other tests with the same mocking.
Also it's better to have it one place when we will support other platforms (not depend on dpkg).

indy_node/test/upgrade/test_forced_upgrade_if_ordered_and_then_request_received.py
```diff
@@ -32,6 +32,6 @@ def test_forced_upgrade_handled_once_if_ordered_and_then_request_received(
     looper.runFor(waits.expectedUpgradeScheduled())
 
     checkUpgradeScheduled([slow_node], validUpgradeExpForceTrue[VERSION])
-    assert len(list(slow_node.upgrader._actionLog)) == 1
-    assert slow_node.upgrader._actionLog.lastEvent[1] == \
-           UpgradeLog.SCHEDULED
+    assert count_action_log_entries(list(slow_node.upgrader._actionLog),
```
ashcherbakov: It looks like we use the same lambda in all (or most) of the tests. Maybe move it to a helper method (or extend `count_action_log_entries`)?

indy_node/test/upgrade/test_pool_upgrade_same_version_reinstall.py
```diff
@@ -4,16 +4,16 @@
...
+    sdk_ensure_upgrade_sent, lowerVersion
 from indy_common.constants import REINSTALL
 
 
 def test_do_upgrade_to_the_same_version_if_reinstall(looper, tconf, nodeSet,
                                                      validUpgrade, sdk_pool_handle,
                                                      sdk_wallet_trustee):
     upgr1 = deepcopy(validUpgrade)
-    upgr1[VERSION] = codeVersion()
+    upgr1[VERSION] = upgr1[VERSION] = lowerVersion(validUpgrade['version'])
```
ashcherbakov: `upgr1[VERSION]` twice?

indy_client/test/cli/test_pool_upgrade.py
```diff
@@ -100,9 +100,8 @@ def test_force_false_upgrade(
```
ashcherbakov: Please add a test where package name is not specified (a default value should be used)

indy_node/test/upgrade/conftest.py
dsurnin: done

indy_client/test/cli/test_pool_upgrade.py
dsurnin: there are several tests which use send_upgrade_cmd func - it uses default for package parameter

PR #891, opened 2018-08-13
Created the script for building graphs from csv files for logs visual…

NataliaDracheva:
The script requires matplotlib 1.5.1 & pandas 0.17.1.

### DIFF
--- a/scripts/build_graph_from_csv.py
+++ b/scripts/build_graph_from_csv.py
@@ -1,0 +1,72 @@
+from matplotlib import pyplot as plt
+import pandas as pd
+import argparse
+
+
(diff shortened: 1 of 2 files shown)


scripts/build_graph_from_csv.py
```diff
@@ -0,0 +1,72 @@
+from matplotlib import pyplot as plt
```
skhoroshavin: Would be great to add `#! /usr/bin/env python3` as a first line, make file executable and remove `.py` extensions to make it more like other scripts

skhoroshavin:
(ci) ok to test

PR #893, opened 2018-08-14
Fix perf reqs

dsurnin:


ashcherbakov:
It looks like we need to cover the load script by unit tests.

PR #895, opened 2018-08-14
 INDY-1548: Read_ledger can handle plugin ledgers

ArtObr:


scripts/read_ledger
```diff
@@ -89,6 +88,7 @@ def get_ledger_dir(node_name, client_name, network):
 
 def get_storage(type_, ledger_data_dir):
     config = getConfig()
+    postfix = config.transactions_file_base_postfix
```
ashcherbakov: For now it may be better to have a hard-coded constant here, so that we can provide the script as is for debug purposes to test pools without a need to re-install IndyNode.
Add a TODO comment to get rid of the constant.

scripts/read_ledger
```diff
@@ -109,6 +111,20 @@ def get_storage(type_, ledger_data_dir):
                                 read_only=True)
 
 
+def get_additional_storages(ledger_data_dir):
+    config = getConfig()
+    postfix = config.transactions_file_base_postfix
+
+    additional_storages = []
+    dir_names = [name for name in os.listdir(ledger_data_dir)
```
ashcherbakov: Can't we just get the list of `additional_storages` by something like
```
additional_storages = [dir_name[:dir_name.find(postfix) for dir_name in os.listdir(ledger_data_dir) if postfix in dir_name]
```

scripts/read_ledger
ArtObr: ok

scripts/read_ledger
ArtObr: ok

PR #935, opened 2018-09-04
Indy-1644, Indy-1666 Updated spike load script in order to generate growing load.

NataliaDracheva:


hyperledger-github:

ashcherbakov:
(ci) test this please

scripts/performance/perf_spike_load.py
```diff
@@ -1,79 +1,206 @@
...
+parser.add_argument('-g', '--graph', default=False, type=bool, required=False, dest='graph',
+                    help='Build a graph to check if the configured profile is correct')
+
+args = parser.parse_args()
+
+
+def create_output_directory(folder_path):
+    output_folder = ""
+    for folder in folder_path:
+        output_folder = os.path.join(output_folder, folder)
```
dsurnin: I think it is not need to use path.join here since func arg is the folder from config. just expand user is enough.

scripts/performance/config_perf_spike_load.yml
```diff
@@ -1,29 +1,42 @@
...
+  seed: 000000000000000000000000Trustee1
+  num: 1
+  refresh: 60
+  buff_req: 300
+  sep: "|"
+  wallet_key: key
+  mode: p
+  pool_config: ""
+  sync_mode: all
+  out: ""
```
dsurnin: I would suggest to specify some value here, e.g. "output", to send all output from console to file by default

dsurnin:
test this please

dsurnin:
test this please

PR #951, opened 2018-09-24
INDY-1700: Updated indy-plenum dependency

spivachuk:


spivachuk:
Closed this pull request because indy-plenum dependency has already been updated in https://github.com/hyperledger/indy-node/pull/949 to a higher version than in this pull request.

PR #956, opened 2018-09-28
Bump and fix for metrics

ArtObr:


scripts/get_metrics
```diff
@@ -210,7 +210,7 @@ def process_storage(storage, args):
 
     print("Profiling info:")
     for m in MetricsName:
-        if m < MetricsName.NODE_PROD_TIME and \
+        if m > 30000 or m < MetricsName.NODE_PROD_TIME and \
```
skhoroshavin: Better add separate if clause with break

PR #970, opened 2018-10-10
Added script for setting dev dependencies for OSX

faisal00813:
This PR contains a script which installs dev dependencies for OSX

### DIFF
--- a/dev-setup/osx/setup-dev-dependencies.sh
+++ b/dev-setup/osx/setup-dev-dependencies.sh
@@ -1,0 +1,61 @@
+#!/bin/bash
+
+brew update
+
+echo 'Installing libsodium...'
+brew install https://raw.githubusercontent.com/Homebrew/homebrew-core/65effd2b617bade68a8a2c5b39e1c3089cc0e945/Formula/libsodium.rb   
+echo 'Installed libsodium'
+
+echo 'Installing RocksDB...'
+brew install rocksdb
+echo 'Installing RocksDB...'
+
+echo 'Installing Charm Crypto...'
+xcode-select --install
+brew install gmp
+brew install pbc
+pushd /tmp
+git clone https://github.com/JHUISI/charm.git
+pushd charm
+./configure.sh --enable-darwin
+make
+make install
+make test
+popd
+rm -rf charm
+popd
+echo 'Installed Charm Crypto'
+
+echo 'Installing libindy...'
+brew install pkg-config
+brew install automake 
+brew install autoconf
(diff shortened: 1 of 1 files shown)


dev-setup/osx/setup-dev-dependencies.sh
```diff
@@ -0,0 +1,61 @@
+#!/bin/bash
+
+brew update
+
+echo 'Installing libsodium...'
+brew install https://raw.githubusercontent.com/Homebrew/homebrew-core/65effd2b617bade68a8a2c5b39e1c3089cc0e945/Formula/libsodium.rb   
+echo 'Installed libsodium'
+
+echo 'Installing RocksDB...'
```
ashcherbakov: I think it need to install a version of RocksDB Plenum really depends on.

dev-setup/osx/setup-dev-dependencies.sh
```diff
@@ -0,0 +1,61 @@
...
+./configure.sh --enable-darwin
+make
+make install
+make test
+popd
+rm -rf charm
+popd
+echo 'Installed Charm Crypto'
+
+echo 'Installing libindy...'
```
ashcherbakov: I think it needs to install a version of libindy Plenum really depends on.

dev-setup/osx/setup-dev-dependencies.sh
```diff
@@ -0,0 +1,61 @@
...
+git clone https://github.com/hyperledger/indy-sdk.git
+pushd indy-sdk/libindy
+cargo build --release
+cp target/release/libindy.dylib /usr/local/lib/
+popd
+rm -rf indy-sdk
+popd
+echo 'Installed libindy'
+
+echo 'Installing libcrypto...'
```
ashcherbakov: I think it needs to install a version of libindy-crypto Plenum really depends on.

faisal00813: In plenum code base the pinned version for the python wrapper seems to be 0.4.3 but the stable version which is available in indy-crypto seems to be only 0.4.2
Is 0.4.2 binary compatible with 0.4.3 wrapper? Can we pin the binary in this script to 0.4.2 of indy-crypto?

ashcherbakov: The correct version is 0.4.5 now (it's just fixed for stable, and will be fixed for master soon).
Ideally it would be great to get this version from setup.py file in Plenum.

PR #973, opened 2018-10-11
Rc 1.6.75

anikitinDSR:


indy_node/utils/node_control_utils.py
```diff
@@ -75,22 +76,25 @@ def update_package_cache(cls):
...
-        return ret
+    def get_deps_tree(cls, package, include, depth=0):
+        if depth < MAX_DEPS_DEPTH:
+            package_info = cls._get_info_from_package_manager(package)
+            ret = [package]
+            deps = []
+            deps_deps = []
+            for dep in include:
+                if dep in package_info:
+                    match = re.search('.*{} \(= ([0-9]+\.[0-9]+\.[0-9]+[\.\+\~0-9A-Za-z]*)\).*'.format(dep), package_info)
```
ashcherbakov: We should probably support `-` as a possible char in version as well

PR #991, opened 2018-10-18
Fix incorrect shell-bang and argparse action

mgbailey:
Signed-off-by: Mike Bailey <mike.bailey@evernym.com>


hyperledger-github:

PR #995, opened 2018-10-23
[pool automation] Molecule docker driver

andkononykhin:
Signed-off-by: Andrey Kononykhin <andkononykhin@gmail.com>


pool_automation/roles/node_install/molecule/docker-ubuntu1604/molecule.yml
```diff
@@ -0,0 +1,35 @@
...
+platforms:
+  - name: instance
+    image: solita/ubuntu-systemd:latest
+    command: /sbin/init
+    capabilities:
+      - SYS_ADMIN
+    volumes:
+      - /sys/fs/cgroup:/sys/fs/cgroup:ro
+    privileged: true
+    network_mode: ${NETWORK_MODE:-bridge}
```
skhoroshavin: Is custom `network_mode` really needed here?

pool_automation/roles/node_install/molecule/docker-ubuntu1604/molecule.yml
```diff
@@ -0,0 +1,35 @@
...
+    config-file: ../common/yamllint
+platforms:
+  - name: instance
+    image: solita/ubuntu-systemd:latest
+    command: /sbin/init
+    capabilities:
+      - SYS_ADMIN
+    volumes:
+      - /sys/fs/cgroup:/sys/fs/cgroup:ro
+    privileged: true
```
skhoroshavin: `privileged` is superset of `SYS_ADMIN` capability so one of them could be dropped. Also it seems like `SYS_ADMIN` is enough, so it's better to prefer `privileged`: https://molecule.readthedocs.io/en/latest/examples.html#systemd-container

pool_automation/roles/node_install/molecule/docker-ubuntu1604/molecule.yml
```diff
@@ -0,0 +1,35 @@
+---
```
skhoroshavin: Given that vagrant is so much slower than docker and cannot be run inside VMs it think that making docker scenario default would be better

pool_automation/roles/node_install/molecule/docker-ubuntu1604/molecule.yml
andkononykhin: Yes, necessary in some circumstances (at least in my dev environment). Will think about that more.

pool_automation/roles/node_install/molecule/docker-ubuntu1604/molecule.yml
andkononykhin: ok

pool_automation/roles/node_install/molecule/docker-ubuntu1604/molecule.yml
andkononykhin: Yes, makes sense

andkononykhin:
test this please

PR #1005, opened 2018-10-31
[INDY-1432] Added a request type for config ledger.

NataliaDracheva:
Signed-off-by: NataliaDracheva <natalia.dracheva@dsr-corporation.com>

### DIFF
--- a/scripts/performance/perf_load/perf_gen_req_parser.py
+++ b/scripts/performance/perf_load/perf_gen_req_parser.py
@@ -6,6 +6,7 @@ from perf_load.perf_req_gen_attrib import RGAttrib, RGGetAttrib
 from perf_load.perf_req_gen_definition import RGGetDefinition, RGDefinition
 from perf_load.perf_req_gen_revoc import RGDefRevoc, RGGetDefRevoc, RGEntryRevoc, RGGetEntryRevoc, RGGetRevocRegDelta
 from perf_load.perf_req_gen_payment import RGGetPaymentSources, RGPayment, RGVerifyPayment
+from perf_load.perf_req_gen_config_ledger import RGConfig
 
 
 class ReqTypeParser:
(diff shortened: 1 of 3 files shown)


hyperledger-github:

skhoroshavin:
(ci) ok to test

scripts/performance/perf_load/perf_gen_req_parser.py
```diff
@@ -14,7 +15,7 @@ class ReqTypeParser:
          "revoc_reg_entry": RGEntryRevoc, "get_nym": RGGetNym, "get_attrib": RGGetAttrib, "get_schema": RGGetSchema,
          "get_cred_def": RGGetDefinition, "get_revoc_reg_def": RGGetDefRevoc, "get_revoc_reg": RGGetEntryRevoc,
          "get_revoc_reg_delta": RGGetRevocRegDelta, "get_payment_sources": RGGetPaymentSources, "payment": RGPayment,
-         "verify_payment": RGVerifyPayment}
+         "verify_payment": RGVerifyPayment, "config": RGConfig}
```
dsurnin: Name "config" is too general. Rename it to something like "pool_cfg_writes" and RGConfig rename accordingly.

scripts/performance/perf_load/perf_req_gen_config_ledger.py
```diff
@@ -0,0 +1,17 @@
...
+from indy import ledger
+
+from perf_load.perf_utils import rawToFriendly
+from perf_load.perf_req_gen import RequestGenerator
+
+
+class RGConfig(RequestGenerator):
+    _req_types = ["111"]
+
+    def _rand_data(self):
```
dsurnin: this functions generates random data for the req. For config type it is enough to generate just random int like this
def _rand_data(self):
      return str(random.randint(0, 99999999))

NataliaDracheva: Fixed.

PR #1007, opened 2018-11-01
[INDY-1792][INDY-1784][pool automation] Script to create an inventory directory with group_vars

andkononykhin:
The script:
- creates inventory directory
- explores available roles for defaults
- provides API to pass custom values for variables
- dumps config files per role as group_vars

Inventory's group_vars might be passed to molecule.

Additionally the PR fixes AWS resources cleanup routine:

- waiting until instance termination complete before removal of the linked security group


andkononykhin:
test this please

pool_automation/roles/aws_manage/defaults/main.yml
```diff
@@ -30,3 +24,12 @@ aws_regions:
   - us-east-2
   - us-west-1
   - us-west-2
+
+instance_count: 4
+
+pa_aws_prefix: PA
```
skhoroshavin: AFAIK this is going to add another tag to EC2 instances, so probably this better be named as `aws_tag_prefix` to be consistent with `aws_tag_namespace` and `aws_tag_role`. Also `prefix` looks very general here, probably something like `owner` or `dept` or `purpose` would be better.

pool_automation/scripts/README.md
```diff
@@ -0,0 +1,55 @@
...
+inventory-dir/
+└── group_vars
+|   └── all
+|       ├── <role-name1>_config.yml
+|       ...
+└── localhost.yml
+```
+
+So it is possible to place your inventory file(s) here (e.g. `inventory-dir/hosts`)
+and pass either the whole directory or an inventory file to [Ansible command line tools][2aceed7f].
```
skhoroshavin: In fact now passing the whole dir as inventory still has quirks because after running `pool_create` keys and ssh_config are also treated as host list, giving lots of warnings.

pool_automation/scripts/inventory-init.py
```diff
@@ -0,0 +1,184 @@
...
+            continue
+
+        # construct user specified config parameters
+        config = {}
+        for _p in params['defaults'].keys():
+            _arg_name = "{}.{}".format(role, _p)
+            if args.get(_arg_name):
+                config[_p] = args[_arg_name]
+
+        _dump_config("{}/{}_config.yml".format(group_vars_dir, role), config, params['defaults'])
```
skhoroshavin: Having single config file per group instead of per role might be more manageable, at least now

pool_automation/pool_destroy.yml
```diff
@@ -3,4 +3,6 @@
   gather_facts: false
   hosts: localhost
   roles:
-    - { role: aws_manage, tag_namespace: test, tag_role: node, instance_count: 0 }
+    - role: aws_manage
+      vars:
+        instance_count: 0
```
skhoroshavin: This can be simplified to
```
  - role: aws_manage
    instance_count: 0
```

pool_automation/roles/aws_manage/defaults/main.yml
```diff
@@ -1,17 +1,11 @@
 ---
 # AWS primary parameters
-aws_type: t2.micro
-aws_image: ami-027583e616ca104df
+aws_ec2_type: t2.micro
+aws_ami_id: ami-027583e616ca104df
```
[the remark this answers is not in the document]
andkononykhin: Could you explain why?

pool_automation/roles/aws_manage/library/stateful_set.py
```diff
@@ -96,15 +98,22 @@ def get_tag(inst, name):
...
-    valid_region_ids = valid_instances(regions, count)
     hosts = []
+    waiters = []
     changed = False
 
+    valid_region_ids = valid_instances(regions, count)
+
     for region in AWS_REGIONS:
+        _terminated = []
+        _launched = []
```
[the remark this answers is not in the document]
andkononykhin: just to separate them from function's top level ones

pool_automation/roles/aws_manage/library/stateful_set.py
```diff
@@ -96,15 +98,22 @@ def get_tag(inst, name):
 
 
 HostInfo = namedtuple('HostInfo', 'tag_id public_ip user')
+AWSWaiterInfo = namedtuple('AWSWaiterInfo', 'waiter kwargs')
```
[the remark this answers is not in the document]
andkononykhin: I think, yes, we do. [EC2.Waiter API|https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ec2.html#EC2.Waiter.InstanceTerminated.wait] accepts a set of useful arguments (like Filter) that might be utilized in future

pool_automation/roles/aws_manage/defaults/main.yml
andkononykhin: Yes, it is going but not in that PR and it would be likely something like `Project`. It wouldn't be a tag prefix but rather a separate tag itself. Here it is used only for resources where no tags are created (for now) and it's really just a prefix for names.

pool_automation/scripts/README.md
andkononykhin: Fixed

pool_automation/scripts/inventory-init.py
andkononykhin: But it would be less readable since defaults will be accumulated for all roles.

pool_automation/pool_destroy.yml
andkononykhin: Thanks. Fixed

andkononykhin:
@skhoroshavin 

> This was a lot of work, however this still has two serious problems:
>pool_install.yml doesn't work with inventory created with inventory_init.py mypool and updated with >ansible-playbook -i mypool pool_create.yml
There were two reasons:
1. ssh configs polluted inventory directory (as you mentioned). Fixed
2. `pool_install` play used `nodes` group but it didn't have any relations with `pool_create` play vars. Fixed
>it's great that inventory_init.py looks for defaults in roles in very generic way, however it would be much >more convenient if it automatically updated aws_tag_namespace accordingly. Yes, it can be updated >manually or with yet another parameter to inventory_init, but it looks odd and I fear this workflow will be >error-prone

Fixed: namespace now is a basename of the inventory if the latter is defined (default is the same)

pool_automation/roles/aws_manage/library/stateful_set.py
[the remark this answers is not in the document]
andkononykhin: Fixed

pool_automation/scripts/inventory-init.py
andkononykhin: Done

pool_automation/roles/aws_manage/defaults/main.yml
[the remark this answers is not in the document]
andkononykhin: `aws_ami_id` has been removed

andkononykhin:
Test this please

PR #1023, opened 2018-11-08
[Pool Automation][INDY-1784] Improves EC2 launching and termination routine

andkononykhin:
Fixes AWS cleanup routine: waiting for ec2 in stances termination complete before trying security groups removal.

### DIFF
--- a/pool_automation/roles/aws_manage/library/test_stateful_set.py
+++ b/pool_automation/roles/aws_manage/library/test_stateful_set.py
@@ -109,14 +109,10 @@ def test_AwsEC2Launcher(ec2):
     instances = launcher.launch(params, 2, ec2=ec2)
 
     assert len(instances) == 2
-    assert len(launcher._ids) > 0
 
-    for instance in instances:
-        instance.reload()
-        assert instance.state['Name'] == 'pending'
-
+    assert len(launcher.awaited) > 0
     launcher.wait()
-    assert len(launcher._ids) == 0
+    assert len(launcher.awaited) == 0
 
     for instance in instances:
         instance.reload()
@@ -133,12 +129,10 @@ def test_AwsEC2Terminator(ec2):
 
     for instance in instances:
         terminator.terminate(instance)
-        instance.reload()
-        assert instance.state['Name'] == 'shutting-down'
 
-    assert len(terminator._ids) > 0
+    assert len(terminator.awaited) > 0
     terminator.wait()
-    assert len(terminator._ids) == 0
+    assert len(terminator.awaited) == 0
 
     for instance in instances:
         instance.reload()
--- a/pool_automation/roles/aws_manage/library/test_stateful_set.py
+++ b/pool_automation/roles/aws_manage/library/test_stateful_set.py
@@ -2,8 +2,11 @@
 import boto3
 import pytest
 
-from stateful_set import AWS_REGIONS, InstanceParams, find_ubuntu_ami, \
-    create_instances, find_instances, valid_instances, get_tag, manage_instances
+from stateful_set import (
+    AWS_REGIONS, InstanceParams, find_ubuntu_ami,
+    AwsEC2Launcher, AwsEC2Terminator, find_instances,
+    valid_instances, get_tag, manage_instances
+)
 
 
 PARAMS = InstanceParams(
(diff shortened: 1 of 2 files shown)


pool_automation/roles/aws_manage/library/test_stateful_set.py
```diff
@@ -93,21 +103,60 @@ def test_find_ubuntu_image(ec2):
...
     params = PARAMS._replace(role='test_create')
-    instances = create_instances(ec2, params, 2)
+    instances = launcher.launch(params, 2, ec2=ec2)
 
     assert len(instances) == 2
+    assert len(launcher._ids) > 0
+
     for instance in instances:
+        instance.reload()
+        assert instance.state['Name'] == 'pending'
```
skhoroshavin: What if AWS is fast enough so state becomes `starting`? In fact I don't this assert is even needed - as far as I can see contract of launcher is:
- on `launch` return asked number of instances without any guarantees about their state
- after `wait` it's guaranteed that all instances are in some defined state

pool_automation/roles/aws_manage/library/test_stateful_set.py
```diff
@@ -93,21 +103,60 @@ def test_find_ubuntu_image(ec2):
...
 
-def test_create_instances(ec2):
+def test_AwsEC2Launcher(ec2):
+    launcher = AwsEC2Launcher()
     params = PARAMS._replace(role='test_create')
-    instances = create_instances(ec2, params, 2)
+    instances = launcher.launch(params, 2, ec2=ec2)
 
     assert len(instances) == 2
+    assert len(launcher._ids) > 0
```
skhoroshavin: Accessing implementation details often leads to fragile tests, is it really worth here?

pool_automation/roles/aws_manage/library/test_stateful_set.py
```diff
@@ -93,21 +103,60 @@ def test_find_ubuntu_image(ec2):
     assert 'UNSUPPORTED' not in image.description
 
 
-def test_create_instances(ec2):
+def test_AwsEC2Launcher(ec2):
+    launcher = AwsEC2Launcher()
     params = PARAMS._replace(role='test_create')
-    instances = create_instances(ec2, params, 2)
+    instances = launcher.launch(params, 2, ec2=ec2)
```
skhoroshavin: I suspect that we'll need separate instances of launcher per region, so it might be more convenient to pass `ec2` to launcher constructor instead of `launch`

pool_automation/roles/aws_manage/library/test_stateful_set.py
```diff
@@ -93,21 +103,60 @@ def test_find_ubuntu_image(ec2):
...
+
     for instance in instances:
+        instance.reload()
+        assert instance.state['Name'] == 'pending'
+
+    launcher.wait()
+    assert len(launcher._ids) == 0
+
+    for instance in instances:
+        instance.reload()
```
skhoroshavin: We already store instances in launcher, can we call `reload` inside `wait` to make API more convenient?

pool_automation/roles/aws_manage/library/test_stateful_set.py
andkononykhin: Why do think so? It would lead to collection of waiters just to serve instance launching across regions. Waiter class can encapsulate that well itself.

pool_automation/roles/aws_manage/library/test_stateful_set.py
andkononykhin: Why do think so? It would lead to collection of waiters just to serve instance launching across regions. Waiter class can encapsulate that well itself.

pool_automation/roles/aws_manage/library/test_stateful_set.py
andkononykhin: Yes. Not actually necessary. Was a kind of debugging

pool_automation/roles/aws_manage/library/test_stateful_set.py
andkononykhin: Waiter uses boto3 client api, reload is from resource api. And in any case we shouldn't rely on implementation details: here we should be sure that instance state is synced before any checks. Thus I don't think it's a good idea.

PR #1034, opened 2018-11-14
INDY-902 test_send_schema

cam-parra:
Signed-off-by: Cam Parra <caeparra@gmail.com>

PR #1040, opened 2018-11-16
WIP INDY-902 send-get-schema

cam-parra:
Signed-off-by: Cam Parra <caeparra@gmail.com>


indy_client/test/cli/test_send_get_schema.py
```diff
@@ -1,112 +1,109 @@
...
+
+def test_send_get_schema_succeeds(looper, sdk_pool_handle,
+                                  sdk_wallet_trust_anchor):
+    _, issuer_id = sdk_wallet_trust_anchor
+
+    schema_name = "Faber"
+    schema_version = "1.8"
+    schema_attr = ["attribute1"]
+
+    write_schema(
```
ArtObr: I think, it would be nice to make a fixture sended_schema (like send_schema in removed tests), instead of writing schema every time

indy_node/test/api/helper.py
```diff
@@ -177,10 +177,42 @@ def validate_claim_def_txn(txn):
 
 # Misc utility
 
-def sdk_write_schema(looper, sdk_pool_handle, sdk_wallet_steward):
+def sdk_write_schema(looper, sdk_pool_handle, sdk_wallet_steward, multi_attribute=[], name="", version=""):
     _, identifier = sdk_wallet_steward
-    _, schema_json = looper.loop.run_until_complete(
-        issuer_create_schema(identifier, "name", "1.0", json.dumps(["first", "last"])))
+
+    if multi_attribute:
```
ArtObr: We should probably write these methods with same style as in https://github.com/hyperledger/indy-node/pull/1034

PR #1041, opened 2018-11-16
[Pool automation][INDY-1626] Client roles and playbooks

andkononykhin:
Adds:
- roles to install indy-cli and perfomance scripts on clients hosts
- improves inventory-init script to support playbook level variables
- improves roles/playbook structure


pool_automation/configure.yml
```diff
@@ -0,0 +1,16 @@
+---
+- name: Configure pool
+  gather_facts: false
+  hosts: nodes
+  roles:
+    - indy_node
+  tags: [ 'pool', 'nodes' ]
```
skhoroshavin: Do we really need `pool` tag here?

andkononykhin: I think it would be handy to be able to refer that using `pool` since the playbook actually configures not a set of separate nodes but the pool of them.

pool_automation/configure.yml
```diff
@@ -1,4 +1,11 @@
 ---
+- name: Check inventory directory
+  gather_facts: false
+  hosts: localhost
+  tasks:
+    - include_tasks: tasks/check_inventory.yml
+      when: not (skip_inventory_check|default(false)|bool)
```
skhoroshavin: This is repeated in 3 different playbooks, probably this better be moved to check_inventory.yml

PR #1044, opened 2018-11-20
Output of shell command were not delivered to the caller - fixed by r…

dsurnin:
…eplacing with proper func call;

### DIFF
--- a/indy_node/utils/node_control_utils.py
+++ b/indy_node/utils/node_control_utils.py
@@ -27,6 +27,9 @@ MAX_DEPS_DEPTH = 6
 
 
 class NodeControlUtil:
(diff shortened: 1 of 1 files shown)


indy_node/utils/node_control_utils.py
```diff
@@ -30,8 +30,7 @@ class NodeControlUtil:
     @classmethod
     def run_shell_command(cls, command, timeout=TIMEOUT):
```
ashcherbakov: Please add comments about the difference and usage for `run_shell_command` and `run_shell_script`.

dsurnin: done

PR #1045, opened 2018-11-20
INDY-902: Integrating tests to SDK 

ArtObr:


indy_client/test/cli/test_send_get_schema.py
```diff
@@ -1,112 +1,112 @@
...
-
-
-def test_send_get_schema_fails_without_dest(
-        be, do, poolNodesStarted, trusteeCli, send_schema):
-
-    with pytest.raises(AssertionError) as excinfo:
-        do('send GET_SCHEMA name=Degree version=1.0',
-            expect=SCHEMA_NOT_FOUND, within=5)
-    assert(INVALID_SYNTAX in str(excinfo.value))
+# import pytest
```
dsurnin: Why do we need whole file commented?

indy_node/test/validator_info/test_validator_info.py
```diff
@@ -102,28 +92,15 @@ def test_validator_info_file_get_claim_def(read_txn_and_get_latest_info,
...
-
-
-def makeGetNymRequest(client, wallet, nym):
+def makeGetNymRequest(looper, sdk_pool_handle, sdk_wallet, nym):
     op = {
         TARGET_NYM: nym,
         TXN_TYPE: GET_NYM,
     }
-    return submitRequests(client, wallet, op)
+    return sdk_submit_operation_and_get_replies(looper, sdk_pool_handle, sdk_wallet, op)
```
dsurnin: here and later
are there any special conditions we do not use dedicated SDKs build_req* functions?

indy_client/test/cli/test_send_get_schema.py
ArtObr: This test will be integrated by Cam in separate PR.

indy_node/test/validator_info/test_validator_info.py
ArtObr: No

PR #1059, opened 2018-11-27
INDY-1770: decrease iptables clients connections limit

sergey-shilov:
Updated docs, migrated indy.env.


docs/setup-iptables.md
```diff
@@ -3,6 +3,11 @@
 In order to prevent the indy-node process from reaching of open file descriptors limit caused by clients connections it is strongly
```
ashcherbakov: What about `CLIENT_CONNECTIONS_LIMIT=10000` in `postinst_node`?

docs/setup-iptables.md
```diff
@@ -3,6 +3,11 @@
...
 recommended to add iptables rule that limits the number of simultaneous clients connections for client port.
```
ashcherbakov: I think a link of this doc and explicit statement about setting firewall rules needs to be part of `README.md` and `start-nodes.md`

docs/setup-iptables.md
sergey-shilov: Sure.

docs/setup-iptables.md
sergey-shilov: Updated _setup-iptables.md_ and _start-nodes.md_.

PR #1063, opened 2018-11-28
Indy 1814

dsurnin:


indy_node/test/validator_info/test_validator_info.py
```diff
@@ -25,19 +24,25 @@ def test_validator_info_file_metrics_count_ledger_field_valid(info):
     assert info['Node_info']['Metrics']['transaction-count']['config'] == 0
 
 
-@pytest.mark.skip(reason="info will not be included by default")
-def test_validator_info_file_software_indy_node_valid(info):
-    assert info['Software']['indy-node'] == node_pgk_version
+def test_validator_info_file_metrics_count_all_ledgers_field_valid(node, info):
```
ashcherbakov: I would add additional ledger here (a plugin?) to check that it's also counted

scripts/validator-info
```diff
@@ -519,7 +527,7 @@ async def handle_client(client_reader, client_writer):
             else:
                 logger.debug("Received data: {}".format(data))
                 stats = json.loads(data.decode())
-                print(json.dumps(stats, indent=2, cls=NewEncoder))
+                print(json.dumps(stats, indent=2, cls=NewEncoder, sort_keys=True))
```
ashcherbakov: I think it's better to use JsonSerializer here

ashcherbakov: Please add a test making sure that the output is sorted

PR #1069, opened 2018-12-01
fix and update acceptance batches and expected results

ozheregelya:


hyperledger-github:

PR #1072, opened 2018-12-04
Old client's lumps removal

ArtObr:


scripts/client_connections/README.md
```diff
@@ -1,38 +0,0 @@
-# Test for checking that all simultaneous clients have a chance to connect to the pool with enabled client stack restart
```
ashcherbakov: This script deals with SDK clients and should not be removed.

scripts/client_connections/just_connect_N_times.py
```diff
@@ -1,121 +0,0 @@
-import argparse
```
ashcherbakov: This script deals with SDK clients and should not be removed.

PR #1074, opened 2018-12-04
[WIP][ci skip][Pool Automation][INDY-1785] Adds support for EC2 spot instances

andkononykhin:
- adds options:
  - to launch spot instances
  - to specify max spot price
- improves tests for `stateful_set` module: 
  - makes more flexible parametrization
  - makes tests scopes/resource independent, thus they might run in parallel (e.g. using `pytest-xdist`)
  - adds option to specify default market type for test ec2 insatnce (`spot`, `on-demand`) from the command line
  - adds tests for spot instances related logic
  - adds more tags for test resources
- improves/fixes `stateful_set`:
  - all generic tags are added at creation time
  - other tags are added only when instances are already running


andkononykhin:
test this please

pool_automation/roles/aws_manage/library/stateful_set.py
```diff
@@ -115,6 +139,12 @@ def __init__(self):
     def terminate(self, instance, region=None):
         instance.terminate()
         self.add_instance(instance, region)
+        # cancel spot request if any
+        if instance.spot_instance_request_id:
+            ec2cl = boto3.client('ec2', region_name=(region if region
```
skhoroshavin: Do we really need low-level client API here?

pool_automation/roles/aws_manage/library/stateful_set.py
```diff
@@ -179,11 +227,12 @@ def _host_info(inst):
 
     valid_region_ids = valid_instances(regions, count)
 
-    for region in AWS_REGIONS:
+    for region in AWS_REGIONS.keys():
```
skhoroshavin: Python iterates dictionaries by keys by default anyway, so there is no real need in specifying `keys` except for additional clarity.

PR #1081, opened 2018-12-06
INDY-1896: allow force upgrade processing in view change

Toktar:


indy_node/test/pool_restart/test_pool_restart.py
```diff
@@ -92,6 +93,22 @@ def test_pool_restart_now_without_datetime(
                      tdir, tconf, START)
 
 
+def test_pool_restart_in_view_change(sdk_pool_handle, sdk_wallet_trustee, looper,
+                                     tdir, tconf, txnPoolNodeSet):
+
+    for node in txnPoolNodeSet:
+        if node.view_changer is None:
```
ashcherbakov: How can view_changer be None here?

PR #1085, opened 2018-12-07
[ci skip][Pool Automation][INDY-1877][1840] Adds tags for volumes and option for volume size

andkononykhin:


andkononykhin:
test this please

PR #1086, opened 2018-12-10
[Pool Automation][INDY-1623] Adds support for plugins installation

andkononykhin:
- `plugins` as a role to manage plugins installation, provides two ways:
  - `plugins_packages` variable as list of packages
  - `plugins_role` variable as a name of separate role for more complex installation scenarios
- improves indy repositories configuration routine to reduce copy-paste


pool_automation/roles/indy_repo/tasks/main.yml
```diff
@@ -0,0 +1,32 @@
+---
+# TODO it might make sense to split the role into indy_node_repo and indy_sdk_repo
```
skhoroshavin: Also it might make sense to move HTTPS support and adding keys to some common role, and put remaining `apt_repository` tasks in corresponding roles (indy_node, indy_cli)

pool_automation/roles/plugins/tasks/main.yml
```diff
@@ -0,0 +1,14 @@
+---
+- name: Execute role {{ plugins_role }}
+  include_role:
+    name: "{{ plugins_role }}"
+  when: not not plugins_role
```
skhoroshavin: Why double `not`?

pool_automation/roles/indy_cli/tasks/main.yml
```diff
@@ -14,27 +14,12 @@
...
-  apt_repository:
-    repo: "deb https://repo.sovrin.org/sdk/deb xenial {{ indy_cli_channel }}"
-    state: present
-  become: true
+- name: Install indy sdk repo
+  import_role:
+    name: indy_repo
+  vars:
+    indy_repo_sdk_channels: "{{ indy_cli_channel }}"
+    indy_repo_sdk: true
```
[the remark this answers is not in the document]
andkononykhin: meta is:
- less explicit
- called before any tasks that is not desired here

pool_automation/roles/indy_node/tasks/main.yml
```diff
@@ -16,27 +16,12 @@
...
-  apt_repository:
-    repo: "deb https://repo.sovrin.org/deb xenial {{ indy_node_channel }}"
-    state: present
-  become: true
+- name: Install indy node repo
+  import_role:
+    name: indy_repo
+  vars:
+    indy_repo_node_channels: "{{ indy_node_channel }}"
+    indy_repo_node: true
```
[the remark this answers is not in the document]
andkononykhin: meta is:
- less explicit
- called before any tasks that is not desired here

pool_automation/roles/indy_repo/tasks/main.yml
andkononykhin: yes, splitting the code will definitely lead to such changes

pool_automation/roles/plugins/tasks/main.yml
andkononykhin: for string->boolean coercion. without that you'll get something like `'plugins_role' is undefined`

pool_automation/roles/indy_repo/defaults/main.yml
```diff
@@ -0,0 +1,7 @@
+---
+indy_repo_node: false
+indy_repo_node_channels: master
```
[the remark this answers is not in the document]
andkononykhin: yes, makes sense to remove them

pool_automation/roles/indy_repo/tasks/main.yml
andkononykhin: on the other hand it might not: https and the key are common for both repos only for the current moment but things might change since they are actually separate repos

pool_automation/roles/plugins/tasks/main.yml
andkononykhin: updated

pool_automation/roles/indy_repo/defaults/main.yml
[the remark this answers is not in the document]
andkononykhin: done

PR #1100, opened 2018-12-18
INDY-1730: Move authorization to new approach

sergey-shilov:
This is the continuation of #1094.


indy_node/server/action_req_handler.py
```diff
@@ -30,39 +30,38 @@ def __init__(self, idrCache: IdrCache,
...
         pass
 
     def validate(self, req: Request):
-        status = None
         operation = req.operation
         typ = operation.get(TXN_TYPE)
         if typ not in self.operation_types:
             return
         origin = req.identifier
         try:
```
ashcherbakov: Do we still need this try-except? This looks like a generic logic for checking if there is a role. It needs to be part of `write_req_validator.validate`

indy_node/server/domain_req_handler.py
```diff
@@ -180,31 +187,22 @@ def _validateNym(self, req: Request):
...
     def _validateExistingNym(self, req: Request, op, originRole, nymData):
         unauthorized = False
-        reason = None
         origin = req.identifier
         owner = self.idrCache.getOwnerFor(op[TARGET_NYM], isCommitted=False)
-        isOwner = origin == owner
+        is_owner = origin == owner
 
-        if not originRole == TRUSTEE and not isOwner:
+        if not originRole == TRUSTEE and not is_owner:
```
ashcherbakov: This check needs to be part of common auth map rules and validator

indy_node/test/nym_txn/test_nym.py
```diff
@@ -11,7 +11,7 @@ def test_non_steward_cannot_create_trust_anchor(
     sdk_wallet_client = sdk_add_new_nym(looper, sdk_pool_handle, sdk_wallet_steward)
     with pytest.raises(RequestRejectedException) as e:
         sdk_add_new_nym(looper, sdk_pool_handle, sdk_wallet_client, role=TRUST_ANCHOR_STRING)
-    e.match('None role cannot')
+    e.match('There is no accepted constraint')
```
ashcherbakov: It would be great to have more descriptive exception than `There is no accepted constraint` for this kind of cases.

indy_common/authorize/auth_request_validator.py
```diff
@@ -21,11 +22,13 @@ def validate(self, request, action_list: [AbstractAuthAction]):
 
 
 class WriteRequestValidator(AbstractRequestValidator, CompositeAuthorizer):
-    def __init__(self, config, auth_map, cache):
+    def __init__(self, config, auth_map, cache, anyone_can_write_map=None):
```
ashcherbakov: I'm not sure that we need `anyone_can_write_map` anymore with the new approach.
Our new auth constraints assume to have the following format:
`(TRUST_ANCHOR, fees=0 ) OR (ANYONE, fees=10).`
So, by default a txn will succeed only when the first constraint is True.
Plugins can register a validator for fees, so that the can validate that the second condition is true.

indy_node/server/domain_req_handler.py
sergey-shilov: Done.

indy_node/server/action_req_handler.py
sergey-shilov: Done.

PR #1112, opened 2018-12-24
INDY-1730: remove role checking in specific handlers.

sergey-shilov:

### DIFF
--- a/indy_common/authorize/authorizer.py
+++ b/indy_common/authorize/authorizer.py
@@ -52,8 +52,8 @@ class RolesAuthorizer(AbstractAuthorizer):
 
     def is_role_accepted(self, request: Request, auth_constraint: AuthConstraint):
         role = self.get_role(request)
-        if role is not None:
-            return role == auth_constraint.role or auth_constraint.role == '*'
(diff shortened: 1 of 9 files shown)


indy_common/authorize/authorizer.py
```diff
@@ -66,7 +66,10 @@ def authorize(self,
                   request: Request,
                   auth_constraint: AuthConstraint,
                   auth_action: AbstractAuthAction=None):
-        if not self.is_role_accepted(request, auth_constraint):
+        is_role_accepted = self.is_role_accepted(request, auth_constraint)
+        if is_role_accepted is None:
+            return False, "role is not found"
```
ashcherbakov: The method can return None if the role is None. But this can be a valid role, if auth constraint has `*` as a role.
So, `is_role_accepted` can return two possible errors:
- did doesn't exist on the ledger (if there is a KeyError in `get_role`)
- role X can not do the operation (if role doesn't match the auth constraint, regardless of whether the role is None or not)

sergey-shilov: 'None' roles are stored as empty strings, so the role returned as None means that sender's DID is not found in a ledger. Left a comment regarding this.

PR #1116, opened 2018-12-29
INDY-1918: Fix for upgrade request

ArtObr:

### DIFF
--- a/indy_node/server/upgrader.py
+++ b/indy_node/server/upgrader.py
@@ -375,7 +375,6 @@ class Upgrader(NodeMaintainer):
                                 upgrade_id=upgrade_id,
                                 reason="problems in communication with node control service")
             self._unscheduleAction()
-            self._actionFailedCallback()
         else:
             logger.info("Waiting {} minutes for upgrade to be performed".format(failTimeout))
             timesUp = partial(self._declareTimeoutExceeded, when, version, upgrade_id)


indy_node/test/upgrade/test_broken_connection_control_tool.py
```diff
@@ -0,0 +1,159 @@
...
+    old_get_version = deepcopy(Upgrader.getVersion)
+    old_update = deepcopy(Upgrader._update_action_log_for_started_action)
+    Upgrader.getVersion = lambda self, pkg: '1.6.0'
+    Upgrader._update_action_log_for_started_action = lambda self: 1
+    yield
+    Upgrader.getVersion = staticmethod(old_get_version)
+    Upgrader._update_action_log_for_started_action = old_update
+
+
+async def _sendUpgradeRequest(self, when, version, upgrade_id, failTimeout, pkg_name):
```
ashcherbakov: Why do we need so many overrides and patches? Can't we just have a more simple test similar to other tests in `upgrade` folder? For example, `test_upgrade_does_not_get_into_loop`?

ArtObr: fixed

indy_node/test/upgrade/test_broken_connection_control_tool.py
```diff
@@ -0,0 +1,69 @@
...
+@pytest.fixture(scope="module")
+def tconf(tconf):
+    old_delta = tconf.MinSepBetweenNodeUpgrades
+    tconf.MinSepBetweenNodeUpgrades = delta
+    yield tconf
+    tconf.MinSepBetweenNodeUpgrades = old_delta
+
+
+@pytest.fixture(scope='function')
+def validUpgrade(nodeIds, tconf, monkeypatch):
```
ashcherbakov: Why can't we use existing `validUpgrade` fixture and update its output if needed?

indy_node/test/upgrade/test_broken_connection_control_tool.py
```diff
@@ -0,0 +1,69 @@
...
+                                   skip_functions):
+    # Emulating connection problems
+    for node in nodeSet:
+        node.upgrader.retry_limit = 0
+
+    # Send upgrade
+    sdk_ensure_upgrade_sent(looper, sdk_pool_handle,
+                            sdk_wallet_trustee, validUpgrade)
+    looper.runFor(len(nodeIds) * delta)
+
```
ashcherbakov: Can we check that Upgrade actually didn't happen in this case, that is we are still on the same version?

indy_node/test/upgrade/test_broken_connection_control_tool.py
ArtObr: existing validUpgrade parametrized with different kind of versions. But probably that doesn't matter. Will change it

PR #1121, opened 2019-01-11
 INDY-1919: Replace platfrom identifying

ArtObr:


setup.py
```diff
@@ -58,7 +58,8 @@
     )],
     install_requires=['indy-plenum-dev==1.6.646',
                       'python-dateutil',
-                      'timeout-decorator==0.4.0'],
+                      'timeout-decorator==0.4.0',
+                      'distro==1.3.0'],
```
ashcherbakov: Since `python-distro` doesn't present in Ubuntu 16.04 canonical repo, we need to build it manually in `build-3rd-parties.sh` (`build_from_pypi distro 1.3.0`)

ArtObr: ok

PR #1145, opened 2019-01-29
[INDY-1963] Make auth error messages more cleanly

anikitinDSR:


indy_common/authorize/auth_constraints.py
```diff
@@ -20,18 +24,30 @@ def __init__(self, role, sig_count, need_to_be_owner=False, metadata={}):
         self.metadata = metadata
         self.constraint_id = ROLE_CONSTRAINT_ID
 
+    def __str__(self):
+        role = get_named_role(self.role) if self.role != '*' else 'ALL'
+        return "Required role: {}, Count of signatures: {}, Need to be owner: {}".format(role,
```
ashcherbakov: I think it's better to say `number of signatures`

ashcherbakov: Maybe we should have shorter description.
Cases:
- owner is None and role != *:
    `X <role> signatures are required `
- owner is None and role == *:
    `X signatures of any role are required`
- owner is not None and number  of sigs = 1:
    `<role> needs to be owner`
- owner is not None, number of sigs = 1, and role == *
    `needs to be owner`
- owner is not None and number  of sigs = 1 and role != *
    `X <role> signatures are required and needs to be owner`
- owner is not None and number  of sigs = 1 and role == *
    `X signatures of any role are required and needs to be owner`
- else (fallback):
    as in the PR

indy_common/authorize/authorizer.py
```diff
@@ -73,11 +77,12 @@ def authorize(self,
         if is_role_accepted is None:
             return False, "sender's DID {} is not found in the Ledger".format(request.identifier)
         if not is_role_accepted:
-            return False, "role is not accepted"
+            return False, "{} can not do this action".format(self.get_named_role_from_req(request))
         if not self.is_sig_count_accepted(request, auth_constraint):
-            return False, "count of signatures is not accepted"
+            return False, "Count of signatures is not accepted"
```
ashcherbakov: Maybe better to say `Not enough signatures`?

indy_common/authorize/authorizer.py
```diff
@@ -73,11 +77,12 @@ def authorize(self,
...
         if not self.is_owner_accepted(auth_constraint, auth_action):
-            return False, "actor must be owner"
+            return False, "{} can not touch verkey field since only the owner can modify it".\
```
ashcherbakov: Why are we talking about verkey field? This is a generic exception, isn't it?

indy_common/authorize/authorizer.py
```diff
@@ -73,11 +79,17 @@ def authorize(self,
...
             return False, "sender's DID {} is not found in the Ledger".format(request.identifier)
         if not is_role_accepted:
-            return False, "role is not accepted"
+            return False, "{} can not do this action".format(self.get_named_role_from_req(request))
         if not self.is_sig_count_accepted(request, auth_constraint):
-            return False, "count of signatures is not accepted"
+            return False, "Not enough signatures"
         if not self.is_owner_accepted(auth_constraint, auth_action):
-            return False, "actor must be owner"
+            if auth_action.txn_type == NYM:
```
ashcherbakov: Maybe we should rather check if this is `AuthActionEdit` action and get the field from there (action.field) instead of having hard-coded `NYM` and `verkey`?

PR #1179, opened 2019-02-20
[INDY-2001] Implement auth rule maps in config ledger

anikitinDSR:

### DIFF
--- a/indy_common/test/auth/test_auth_strategies.py
+++ b/indy_common/test/auth/test_auth_strategies.py
@@ -46,16 +46,6 @@ def is_accepted():
 
 
 @pytest.fixture
-def state():
-    return PruningState(KeyValueStorageInMemory())
-
-
-@pytest.fixture
-def state_serializer():
-    return ConstraintsSerializer(domain_state_serializer)
-
-
-@pytest.fixture
 def config_ledger_strategy(state, state_serializer):
     return ConfigLedgerAuthStrategy(auth_map,
                                     state=state,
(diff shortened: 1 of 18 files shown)


indy_node/server/node.py
```diff
@@ -361,3 +370,29 @@ def is_request_need_quorum(self, msg_dict: dict):
...
+        for rule_id, auth_constraint in auth_map.items():
+            serialized_key = rule_id.encode()
+            serialized_value = serializer.serialize(auth_constraint)
+            if not state.get(serialized_key, isCommitted=False):
+                state.set(serialized_key, serialized_value)
+
+    def _init_write_request_validator(self):
+        constraint_serializer = ConstraintsSerializer(domain_state_serializer)
+        config_state = self.states[CONFIG_LEDGER_ID]
+        self.add_auth_rules_to_config_state(state=config_state,
```
ashcherbakov: We should not modify state without consensus. If a rule is not present in the config state - use local auth map instead.

indy_common/test/auth/test_auth_strategies.py
```diff
@@ -88,3 +110,48 @@ def test_is_accepted_for_all_new_value(is_accepted):
 def test_is_accepted_for_all(is_accepted):
     assert is_accepted("PREFIX--TYPE--*--*--*",
                        "PREFIX--TYPE--FIELD--OLD_VALUE--NEW_VALUE")
+
+
+def test_config_strategy_get_constraint_from_state(state,
```
ashcherbakov: Please move config stategy tests into a separate file

PR #1187, opened 2019-02-28
[INDY-2006] add updateState into config request handler

anikitinDSR:


indy_node/test/auth_rule_change/test_catching_up_auth_rule_txn.py
```diff
@@ -0,0 +1,45 @@
...
+                                             constraint=changed_constraint.as_dict)
+        delayed_node.start_catchup()
+        looper.run(eventually(lambda: assertExp(delayed_node.mode == Mode.participating)))
+    sdk_add_new_nym(looper,
+                    sdk_pool_handle,
+                    sdk_wallet_steward,
+                    'newSteward2',
+                    STEWARD_STRING,
+                    dest=new_steward_did, verkey=new_steward_verkey)
+
```
ashcherbakov: I think it makes sense to check here that the new NYM was added on all nodes including `delayed_node`, that is call `ensure_all_nodes_have_same_data`

indy_node/test/auth_rule_change/test_catching_up_auth_rule_txn.py
```diff
@@ -0,0 +1,45 @@
...
+    delayed_node = txnPoolNodeSet[-1]
+    wh, _ = sdk_wallet_trustee
+    new_steward_did, new_steward_verkey = create_verkey_did(looper, wh)
+    changed_constraint = AuthConstraint(role=STEWARD,
+                                        sig_count=1)
+    action = AuthActionAdd(txn_type=NYM,
+                           field=ROLE,
+                           value=STEWARD)
+    with delay_rules_without_processing(delayed_node.nodeIbStasher, cDelay(), pDelay(), ppDelay()):
+        sdk_send_and_check_auth_rule_request(looper, sdk_wallet_trustee,
```
ashcherbakov: Does it make sense to check that 1 steward can not add another Steward before this is changed? Otherwise the test may give false positive results in future if we change this default rule.

indy_node/server/config_req_handler.py
```diff
@@ -202,3 +206,20 @@ def get_auth_key(operation):
...
+
+    @staticmethod
+    def get_auth_constraint(operation):
+        return ConstraintCreator.create_constraint(operation.get(CONSTRAINT))
+
+    def update_auth_constraint(self, auth_key, constraint):
+        self.state.set(auth_key.encode(),
+                       self.constraint_serializer.serialize(constraint))
+
+    def updateState(self, txns, isCommitted=False):
```
ashcherbakov: I think we need to integrate it for the new Pluggable Request Handler for AUTH_RULE txn.

PR #1207, opened 2019-03-14
Update transaction structure description.

Diiaablo95:
The description of the content of a transaction, and in particular a node transaction, is not consistent with the current workings of indy-node. While in the [transactions README](https://github.com/hyperledger/indy-node/blob/master/docs/source/transactions.md) is stated that the `dest` field of a node transaction is a node DID, it is instead the node verification key (as also shown by the test scripts, where the verification key of the node is written in the pool transaction, and not it public key).

### DIFF
--- a/docs/source/transactions.md
+++ b/docs/source/transactions.md
@@ -458,16 +458,10 @@ Adds a new node to the pool or updates an existing node in the pool
 
 - `dest` (base58-encoded string):
 
-    Target Node's DID as base58-encoded string for 16 or 32 byte DID value.
+    Target Node's verkey as base58-encoded string for 16 or 32 byte DID value.
     It differs from `identifier` metadata field, where `identifier` is the DID of the transaction submitter (Steward's DID).
 
-    *Example*: `identifier` is a DID of a Steward creating a new Node, and `dest` is the DID of this Node.
-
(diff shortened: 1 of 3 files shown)


hyperledger-github:

docs/source/transactions.md
```diff
@@ -458,16 +458,10 @@ Adds a new node to the pool or updates an existing node in the pool
 
 - `dest` (base58-encoded string):
 
-    Target Node's DID as base58-encoded string for 16 or 32 byte DID value.
+    Target Node's verkey as base58-encoded string for 16 or 32 byte DID value.
```
ashcherbakov: Yes,  you are right. abbreviated verkeys are currently not supported. Can you please also fix the documentation for NODE request in `requests.md`?

ashcherbakov:
(ci) test this please

docs/source/transactions.md
Diiaablo95: Hey! Just updated the same info in both `requests.md` and `requests-new.md`.

ashcherbakov:
(ci) test this please

docs/source/transactions.md
ashcherbakov: Thank you very much!

PR #1213, opened 2019-03-20
INDY-2018: Node's verkey validation fix

ArtObr:


indy_node/server/pool_req_handler.py
```diff
@@ -34,6 +38,13 @@ def authErrorWhileAddingNode(self, request):
         if error:
             return error
 
+        dest = operation.get(TARGET_NYM)
```
ashcherbakov: Should we check it when we edit NODE txn (changing TARGET_NYM)?

indy_node/server/pool_req_handler.py
```diff
@@ -34,6 +38,13 @@ def authErrorWhileAddingNode(self, request):
...
+        did = base58.b58decode(dest)
```
ashcherbakov: Can we use some helper method from (like `ed25519PkToCurve25519`) for this?

ArtObr: yes, will be fixed.

indy_node/server/pool_req_handler.py
ArtObr: Will be extended for verkey changing

indy_node/server/pool_req_handler.py
```diff
@@ -34,6 +39,15 @@ def authErrorWhileAddingNode(self, request):
         if error:
             return error
 
+        dest = operation.get(TARGET_NYM)
```
ashcherbakov: Actually it can be done as part of static validation, not dynamic one. See `doStaticValidation`.

indy_node/test/node_txn/test_send_node_with_invalid_verkey.py
```diff
@@ -0,0 +1,74 @@
...
+                             services=[VALIDATOR],
+                             key_proof=key_proof))
+
+    node_request = json.loads(node_request)
+    node_request['operation'][VERKEY] = invalid_dest
+    node_request = json.dumps(node_request)
+
+    request_couple = sdk_sign_and_send_prepared_request(looper, new_steward_wallet_handle,
+                                                        sdk_pool_handle, node_request)
+    sdk_get_bad_response(looper, [request_couple], RequestRejectedException,
```
ashcherbakov: It the new logic is part of static validation, then it will throw NACK instead of Reject

indy_node/server/pool_req_handler.py
ArtObr: ok

indy_node/test/node_txn/test_send_node_with_invalid_verkey.py
ArtObr: ok

PR #1243, opened 2019-04-17
INDY-2043: add AuthRuleValueField and change an auth key format

Toktar:

### DIFF
--- a/indy_common/authorize/authorizer.py
+++ b/indy_common/authorize/authorizer.py
@@ -164,7 +164,7 @@ class OrAuthorizer(AbstractAuthorizer):
                                       auth_constraint=constraint,
                                       auth_action=auth_action)
             except AuthValidationError as e:
-                logger.debug(e)
+                logger.trace(e)
             else:
                 successes.append(True)
         if len(successes) == 0:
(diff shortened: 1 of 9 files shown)


indy_common/authorize/authorizer.py
```diff
@@ -162,7 +164,7 @@ def authorize(self,
                                       auth_constraint=constraint,
                                       auth_action=auth_action)
             except AuthValidationError as e:
-                print(e)
+                logger.debug(e)
```
ashcherbakov: Maybe `trace`?

indy_common/test/types/test_get_auth_rule_schema.py
```diff
@@ -22,3 +22,11 @@ def test_has_expected_validators():
     schema = dict(ClientGetAuthRuleOperation.schema)
     for field, validator in EXPECTED_ORDERED_FIELDS.items():
         assert isinstance(schema[field], validator)
+
+
```
ashcherbakov: I think we need to mention `AuthRuleValueField` in `EXPECTED_ORDERED_FIELDS ` above as a value for `"old_value"` and `"new_value"`

PR #1246, opened 2019-04-19
Attrib auth rule fix

ArtObr:


indy_node/server/domain_req_handler.py
```diff
@@ -210,14 +212,33 @@ def _validateAttrib(self, req: Request):
...
 
-        if op.get(TARGET_NYM) and op[TARGET_NYM] != req.identifier and \
-                not self.idrCache.getOwnerFor(op[TARGET_NYM],
-                                              isCommitted=False) == origin:
-            raise UnauthorizedClientRequest(
-                req.identifier,
-                req.reqId,
-                "Only identity owner/guardian can add attribute "
-                "for that identity")
+        is_owner = self.idrCache.getOwnerFor(op[TARGET_NYM],
```
ashcherbakov: It looks like `TARGET_NYM` is an optional field, while in `prepare_attr_for_state` and validation we assume it may be not.
I suggest to remove `optional` tagfrom Atrrib's `TARGET_NYM` (see `ClientAttribOperation`)

ArtObr: agreed

ArtObr:
(ci) test this please

indy_node/server/domain_req_handler.py
```diff
@@ -210,14 +213,52 @@ def _validateAttrib(self, req: Request):
...
+                field = key
+                value = op[key]
+                break
+        if field is None or value is None:
+            raise LogicError('Attribute data cannot be empty')
+
+        get_key = None
+        if field == RAW:
+            try:
+                get_key = json.loads(value)
```
ashcherbakov: It will be better to use `attrib_raw_data_serializer` instead of `json`.

indy_node/server/domain_req_handler.py
```diff
@@ -210,14 +213,52 @@ def _validateAttrib(self, req: Request):
...
+                raise InvalidClientRequest(origin, req.reqId,
+                                           'Attribute field must be dict while adding it as a row field'.
+                                           format(TARGET_NYM))
+        else:
+            get_key = value
+
+        if get_key is None:
+            raise LogicError('Attribute data must be parsed')
+
+        old_value, seq_no, _, _ = self.getAttr(op[TARGET_NYM], get_key, field)
```
ashcherbakov: I think we need to call `getAttr` with `isCommitted=False`

indy_node/test/attrib_txn/test_nym_attrib.py
```diff
@@ -251,3 +276,68 @@ def testTrustAnchorAddedAttributeCanBeChanged(sdk_added_raw_attribute):
     # TODO but only by user(if user has taken control of his identity) and
     # trustAnchor
     raise NotImplementedError
+
+
+def test_auth_rule_for_attrib_works(looper,
```
ashcherbakov: This test is for `raw` attribute. It would be great to have similar tests for `hash` and `enc` attributes as well.

indy_node/server/domain_req_handler.py
ArtObr: ok

indy_node/server/domain_req_handler.py
ArtObr: ok

indy_node/test/attrib_txn/test_nym_attrib.py
ArtObr: ok

PR #1274, opened 2019-04-30
[WIP] INDY-2066: Implement setting up Transaction Author Agreement

skhoroshavin:


indy_node/server/config_req_handler.py
```diff
@@ -167,6 +175,16 @@ def validate(self, req: Request):
                                                               old_value="*",
                                                               new_value="*")])
 
+        elif typ == TXN_ATHR_AGRMT:
+            version = operation[TXN_ATHR_AGRMT_VERSION]
+            if self.state.get(":taa:v:{}".format(version), isCommitted=False) is not None:
```
ashcherbakov: I would use helper methods (`indy_common/state/config.py`, see also `indy_common/state/domain.py`) to access state keys.

indy_node/server/config_req_handler.py
```diff
@@ -277,3 +297,11 @@ def _check_auth_key(self, operation, identifier, req_id):
             raise InvalidClientRequest(identifier, req_id,
                                        "Unknown authorization rule: key '{}' is not "
                                        "found in authorization map.".format(auth_key))
+
+    def update_txn_athr_agrmt(self, payload: dict):
+        version = payload[TXN_ATHR_AGRMT_VERSION]
+        text = payload[TXN_ATHR_AGRMT_TEXT]
+        hash = sha256('{}{}'.format(version, text).encode()).digest()
```
ashcherbakov: `hash` is a standard function, so it's better to find another name for the variable.

PR #1275, opened 2019-05-06
Updates client taa acceptance fields

andkononykhin:


skhoroshavin:
Additional proposal: what about changing `ATHR -> AUTHOR` and `AGRMT -> AGREEMENT` in constant names? Reasons why I think this would make design better:
- these abbreviations are not widespread, are not intuitive (for example why AGRMT and not AGMNT or AGRMNT?) and can lead to confusion, especially when searching through source code
- these abbreviations save only 6 letters in 15+ letters strings (which make them 20+ letters), which I believe is not good enough to justify above mentioned downside
- there is a [PR](https://github.com/hyperledger/indy-sdk/pull/1602/files) to Indy SDK in which nonabbreviated terms are used in many places and IMO they look much cleaner than those abbreviations

design/txn_author_agreement.md
```diff
@@ -83,27 +83,27 @@ Add new version of AML.
...
     "identifier": "<str ident>",
     "protocolVersion": INT,
-    "signature": "<signature (txnAuthrAgrmtMeta is also signed)>",
-    "txnAuthrAgrmtMeta": { 
-        "acceptanceMechanismType": "label of type", 
-        "hash": "hash string", 
-        "timeOfAcceptance": <integer UTC TS>
+    "signature": "<signature (taaAcceptance is also signed)>",
+    "taaAcceptance": {
+        "hash": "hash string",
```
skhoroshavin: Another option is to name it `digest`

andkononykhin:
test this please

design/txn_author_agreement.md
```diff
@@ -83,27 +83,27 @@ Add new version of AML.
...
+        "mechanism": "label of type",
+        "time": <integer UTC TS>
```
jovfer: @andkononykhin @ashcherbakov @skhoroshavin  I'm in doubt is just `time` here unambiguous here or not. In general, in TAA flow "time" may mean
- time of signature
- time of acceptance
- time of TAA itself.

In this particular JSON the next level is "taaAcceptance" (not the original name) so I'm OK with this renaming, but not sure is it good idea or not

design/txn_author_agreement.md
```diff
@@ -83,27 +83,27 @@ Add new version of AML.
...
     "operation" { ... },
     "identifier": "<str ident>",
     "protocolVersion": INT,
-    "signature": "<signature (txnAuthrAgrmtMeta is also signed)>",
-    "txnAuthrAgrmtMeta": { 
-        "acceptanceMechanismType": "label of type", 
-        "hash": "hash string", 
-        "timeOfAcceptance": <integer UTC TS>
+    "signature": "<signature (taaAcceptance is also signed)>",
+    "taaAcceptance": {
```
jovfer: This JSON is a part of public API. I don't like the idea to use abbreviation in public API here

ashcherbakov: I don't mind calling Transaction Author Agreement just TAA. 
1) This is not true public API, at least not for the app developer. 
2) If the user doesn't understand what Transaction Author Agreement is, then it doesn't matter how we call it. Otherwise both terms will be clear, especially if we document it properly.
3) The size of request matters.

design/txn_author_agreement.md
skhoroshavin: I think that `time` here is unambigous - it is time of acceptance, since its under `taaAcceptance`. As for other variants:
- it cannot be time of signature or transaction since it is under very specific `taaAcceptance` section, and if we will need time of signature or transaction we definitely won't add them under this section - they just don't belong here
- it cannot be time of TAA itself because it is already contained in metadata of previously written TAA transaction

design/txn_author_agreement.md
skhoroshavin: When considering this design `TAA` abbreviation comes up numerous times, furthermore - we have request format documentation, where it can be (and should be) described as well.

design/txn_author_agreement.md
```diff
@@ -83,27 +83,27 @@ Add new version of AML.
 Fetch AML from the ledger valid for specified time or the latest one.
 
 ### TAA Verification
-If TAA enabled on the ledger, then each write request from the user must contain metadata about a TAA signed by the user. The new format of write request is 
+If TAA enabled on the ledger, then each write request from the user must contain TAA acceptance signed by the user. The new format of write request is
```
andkononykhin: `acceptance data`

PR #1281, opened 2019-05-08
INDY-2066: Bump plenum version + TAA-related fixups

skhoroshavin:


indy_node/test/auth_rule/test_auth_map.py
```diff
@@ -30,6 +30,14 @@ def test_auth_map_nym():
         assert rule_str in auth_map.auth_map.keys()
```
ashcherbakov: Please add the following tests as well:
- a unit test to `indy_common/test/auth` (`test_txn_author_agreement_with_new_auth_map.py`)
- an integration test for auth rule changing into `indy_node/test/auth_rule/auth_framework`

skhoroshavin: Yes, I have these planned for next PRs

PR #1295, opened 2019-05-21
INDY-2102: Change GET_AUTH_RULE reply format

Toktar:


Toktar:
test this please

indy_node/test/auth_rule/test_get_auth_rule.py
```diff
@@ -143,12 +135,58 @@ def test_get_all_auth_rule_transactions_after_write(looper,
...
+                                           sdk_pool_handle)
+
+    # prepare action key
+    dict_key = dict(auth_rules[0][1][RESULT][DATA][0])
+    dict_key.pop(CONSTRAINT)
+
+    # prepare "operation" to send AUTH_RULE txn
+    op = dict(auth_rules[0][1][RESULT][DATA][0])
+    op[TXN_TYPE] = AUTH_RULE
+    op[CONSTRAINT] = constraint
```
ashcherbakov: Can we send AUTH_RULE with the same Constraint (that it don't override ` op[CONSTRAINT] `)?

Toktar: Yes. Are we need to add this test?

ashcherbakov: Yes, would be great, to make sure that we can use the GET output as-is for the SET input

PR #1301, opened 2019-05-22
INDY-2090: Support TAA in load script

skhoroshavin:


scripts/performance/perf_load/perf_processes.py
```diff
@@ -95,13 +95,20 @@
 parser.add_argument('--test_conn', action='store_true', dest='test_conn',
                     help='Check pool connection with provided genesis file')
 
+parser.add_argument('--taa_text', default="", type=str, required=False,
```
ashcherbakov: Maybe it would be more convenient for our use cases to have a default TAA?

scripts/performance/perf_load/perf_client.py
```diff
@@ -104,6 +110,67 @@ def __init__(self, name, pipe_conn, batch_size, batch_rate, req_kind, buff_req,
...
+            # We reached desired state
+            if current_text == text and current_version == version:
+                self._taa_text = current_text
+                self._taa_version = current_version
+                self._taa_time = current_time + 1  # We are "signing" just 1 second after TAA created
+                break
+
+            # Try to set taa
+            set_taa = await ledger.build_txn_author_agreement_request(self._test_did, text, version)
+            await ledger.sign_and_submit_request(self._pool_handle, self._wallet_handle, self._test_did, set_taa)
```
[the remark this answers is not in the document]
skhoroshavin: This already waits for reply, we just don't check its contents

scripts/performance/perf_load/perf_processes.py
skhoroshavin: I'm inclined to agree

scripts/performance/perf_load/perf_client.py
```diff
@@ -236,8 +325,16 @@ def read_cb(self):
...
             self._loop.stop()
             raise e
+
         await self._req_generator.on_request_generated(req_data, sig_req)
 
+    async def append_taa_acceptance(self, req):
+        if self._taa_text == "":
+            return req
+
+        return await ledger.append_txn_author_agreement_acceptance_to_request(
```
ashcherbakov: If this is XFER request, then TAA metadata is added a bit differently.
```
extra = prepare_payment_extra_with_acceptance_data(extra, text, version, None, mechanism, round(time.time()))
indy_build_payment_req = (...., extra=extra)
```

skhoroshavin: What's worse - in case of XFER requests this needs to be done as part of transaction creation, not afterwards. For now I created a dirty fix, will improve in next PR.

PR #1302, opened 2019-05-22
INDY-2102: change auth_map format

Toktar:


indy_common/authorize/auth_map.py
```diff
@@ -249,104 +250,106 @@
      AuthConstraint(STEWARD, 1, need_to_be_owner=True),
      AuthConstraint(TRUST_ANCHOR, 1, need_to_be_owner=True)])
```
ashcherbakov: Please add a test that we expect the order in `GET_AUTH_RULE()`

PR #1305, opened 2019-05-22
[INDY-2096] Usage of libindy API calls to build auth rules set/get requests

andkononykhin:
- use libindy builders API for AUTH_RULE and GET_AUTH_RULE requests everywhere except tests were invalid requests should be constructed and sent to ledger
- removes auth rule for editing SCHEMA from both `auth_map` and `anyone_can_write_auth_map`


andkononykhin:
test this please

indy_node/test/auth_rule/test_get_auth_rule.py
```diff
@@ -154,42 +157,40 @@ def test_get_all_auth_rule_transactions_after_write(looper,
...
+    )
+    auth_rules = auth_rules_resp[0][1][RESULT][DATA]
+
+    for rule in auth_rules:
+        # prepare action key
+        dict_key = dict(rule)
+        dict_key.pop(CONSTRAINT)
+
+        # prepare AUTH_RULE request
+        req_json = build_auth_rule_request_json(
```
ashcherbakov: The goal of the test is to make sure that we can send GET_AUTH_RULE response to AUTH_RULE `as-is`.
But, as I can see, `build_auth_rule_request_json` re-builds the AUTH_RULE json from the dict.
Maybe we can have two tests for GET - SET case: one we send the result as-is, and one when we use `build_auth_rule_request_json`.

indy_node/test/auth_rule/test_get_auth_rule.py
```diff
@@ -198,32 +199,32 @@ def test_auth_rule_after_get_auth_rule(looper,
...
+    dict_key = dict(rule)
     dict_key.pop(CONSTRAINT)
 
-    # prepare "operation" to send AUTH_RULE txn
-    op = dict(auth_rules[0][1][RESULT][DATA][0])
-    op[TXN_TYPE] = AUTH_RULE
-    op[CONSTRAINT] = constraint
+    # prepare AUTH_RULE request
+    rule[CONSTRAINT] = constraint
+    req_json = build_auth_rule_request_json(
```
ashcherbakov: Here we re-build the request as well, but it would be better to send the result json 'as-is' (just change the constraint)

PR #1306, opened 2019-05-23
INDY-2074: Extend documentation for Auth Rules, TAA and Revocation

ashcherbakov:
- Fix/extend docs for AUTH_RULE and GET_AUTH_RULE
- Add docs about TAA and TAA_AML transactions and requests
- Add docs about Revocation transactions and requests
- Extend Auth Rules with TAA and TAA_AML
- Extend Auth Rules with a description on who is the owner for each action
- Remove anyone_can_write docs from Auth Rules
- Add more tests for error messages when metadata is set

### DIFF
--- a/docs/source/requests.md
+++ b/docs/source/requests.md
@@ -33,7 +33,7 @@
 
 * [Action Requests](#action-requests)
 
-    * [POOL_RESTART](#pool_restrt)
+    * [POOL_RESTART](#pool_restart)
     * [VALIDATOR_INFO](#validator_info)
     
 This doc is about supported client's Request (both write and read ones).
@@ -256,7 +256,7 @@ of a transaction in the Ledger (see [transactions](transactions.md)).
             Unique ID number of the request with transaction.
   
         - `digest` (SHA256 hex digest string):
-            SHA256 hash hex digest of the all fields in the initial requests (including signatures) 
+            SHA256 hash hex digest of all fields in the initial requests (including signatures) 
             
         - `payloadDigest` (SHA256 hex digest string):
             SHA256 hash hex digest of the payload fields in the initial requests, that is all fields excluding signatures and plugins-added ones
@@ -1390,6 +1390,7 @@ ConstraintEntity
 ```
 {
     'operation': {
+        'type':'120',
         'auth_type': '0', 
         'auth_action': 'EDIT',
         'field' :'services',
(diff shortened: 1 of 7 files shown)


docs/source/transactions.md
```diff
@@ -706,35 +899,44 @@ Command to change Pool's configuration
...
+
+The list of actions is static and can be found in [auth_rules.md](auth_rules.md).
+There is a default Auth Constraint for every action (defined in [auth_rules.md](auth_rules.md)). 
+
+The `AUTH_RULE` command allows to change the Auth Constraint.
+So, it's not possible to register new actions by this command. But it's possible to override authentication constraints (values) for a given action.
+
+There are two types of constraints:
+- ConstraintEntity contains `{constraint_id, role, sig_count, need_to_be_owner, metadata}`
+- ConstraintList with format `{constraint_id, auth_constraints}` contains list of constraints.
```
skhoroshavin: I think it would be better to specify possible values of `contraint_id`. Also `ContraintEntity` and `ContraintList` identifiers look a bit artificial here (they are defined, but are not referenced anywhere), so probably something like
```
- constraint_id: ROLE - this is a role constraint, following fields are required...
- constraint_id: OR or AND - this is a list constraint, following fields are required...
```
might look better.

docs/source/transactions.md
```diff
@@ -706,35 +899,44 @@ Command to change Pool's configuration
...
+                            'role': '2',
+                            'sig_count': 1, 
+                            'need_to_be_owner': True, 
+                            'metadata': {}}
+                           ]
+}, 
+
+                                                                
+```
+means that changing a value of a NODE transaction's `service` field from `[VALIDATOR]` to `[]` (demotion of a node) can only be done by two TRUSTEE or one STEWARD who is the owner (the original creator) of this transaction.
```
skhoroshavin: Having reference to mapping between role codes and role names would be nice

docs/source/requests.md
```diff
@@ -8,23 +8,32 @@
...
     * [SCHEMA](#schema)
     * [CLAIM_DEF](#claim_def)
+    * [REVOC_REG_DEF](#revoc_reg_def)
+    * [REVOC_REG_ENTRY](#revoc_reg_entry)
     * [NODE](#node)
     * [POOL_UPGRADE](#pool_upgrade)
     * [POOL_CONFIG](#pool_config)
     * [AUTH_RULE](#auth_rule)
+    * [TRANSACTION_AUTHOR_AGREEMENT](#transaction_author_agreement)
+    * [TRANSACTION_AUTHOR_AGREEMENT_AML](#transaction_author_agreement_AML)
```
skhoroshavin: Current order doesn't match transaction codes, do we care about it?

ashcherbakov: I think we don't care. Here the transactions are sorted just by ledgers.

docs/source/requests.md
```diff
@@ -195,6 +255,22 @@ of a transaction in the Ledger (see [transactions](transactions.md)).
...
+        - `digest` (SHA256 hex digest string):
+            SHA256 hash hex digest of all fields in the initial requests (including signatures) 
+            
+        - `payloadDigest` (SHA256 hex digest string):
+            SHA256 hash hex digest of the payload fields in the initial requests, that is all fields excluding signatures and plugins-added ones
+            
+        - `taaAcceptance` (dict, optional):
+            If transaction author agreement is set/enabled, then every transaction (write request) from Domain and plugins-added ledgers must include acceptance of the latest transaction author agreement.
+            
+                - `taaDigest` (SHA256 hex digest string): SHA256 hex digest of the latest Transaction Author Agreement on the ledger
```
skhoroshavin: I think it should be noted either here or somewhere else in this document that digest is calculated from concatenation of TAA version and text.

ashcherbakov:
> General note on `requests.md` - it would be nice to have documentation for expected keys and values for state proofs in read requests.

Agree. I think it needs to be a separate doc.

PR #1317, opened 2019-05-27
ST-568 Fix Fees

KitHat:


scripts/performance/perf_load/perf_client_fees.py
```diff
@@ -29,6 +31,27 @@ def __init_plugin_once(cls, plugin_lib_name, init_func_name):
                 print("Payment plugin initialization failed: {}".format(repr(ex)))
                 raise ex
 
+    @classmethod
+    async def __set_fees_once(cls, wallet_handle, set_fees, test_did, payment_method, trustee_dids, pool_handle, auth_rule_metadata):
```
ashcherbakov: I think we need to use `static` str prefixes for fees aliases, so that re-sending of SET_FEES (from another client for example) doesn't break anything.

PR #1328, opened 2019-05-29
ST-568: Improve stability and logging of token-related parts of load script

skhoroshavin:

### DIFF
--- a/scripts/performance/perf_load/perf_req_gen_payment.py
+++ b/scripts/performance/perf_load/perf_req_gen_payment.py
@@ -82,11 +82,15 @@ class RGPayment(RGBasePayment):
                 return
 
             try:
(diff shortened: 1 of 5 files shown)


scripts/performance/perf_load/perf_req_gen_payment.py
```diff
@@ -82,11 +82,15 @@ def _gen_req_data(self):
                 return
 
             try:
-                receipt_infos_json = await payment.parse_payment_response(self._payment_method, resp_or_exp)
+                reply = json.loads(resp_or_exp)
```
ashcherbakov: This is generator of XFER requests only, isn't it?
If so, then how can we have anything except XFER Reply here?
As I can see, non-XFER FEEs reply is processed by `_parse_fees_resp` in `LoadClientFees`

PR #1351, opened 2019-06-19
INDY-2144: Test for forced upgrade validation

ArtObr:


ArtObr:
unskip ci when https://github.com/hyperledger/indy-plenum/pull/1243 merged

indy_node/test/upgrade/test_node_sustain_invalid_upgrade_txn.py
```diff
@@ -0,0 +1,20 @@
+import pytest
+
+from plenum.common.exceptions import RequestNackedException
+from plenum.test.pool_transactions.helper import sdk_add_new_nym
+
+from indy_node.test.upgrade.helper import sdk_ensure_upgrade_sent
+
+
+def test_forced_request_validation(looper, txnPoolNodeSet,
```
ashcherbakov: This test checks forced upgrade behavior when nodes get it through Request (that is processRequest method).
Also, we need to check it when the node doesn't get the Request, but receives upgrade as a result of Propagate (that is processPropagate mehod). We may need to delay Requests on some nodes to simulate it. This can be a separate test.

ArtObr:
test this please

indy_node/test/upgrade/test_node_sustain_invalid_upgrade_txn.py
ArtObr: I think we still process propagates, cause part of nodes succesfully validate request and send propagate. Additional checks added

ashcherbakov: Sure, we do process propagates, but we process them AFTER we processed requests. So, we call forced upgrade BEFORE processPropagate method. But we need to test 2 cases: when we call force upgrade via Request (as in this test), and via Propagate.

PR #1356, opened 2019-06-25
[INDY-2151] Nightly builds

andkononykhin:
- adds new pipeline that performs system tests only and might be used in Jenkins jobs triggered periodically
- system testing logic has been moved to local shared groovy module which is consumed by both CD pipeline and nightly build


andkononykhin:
test this please

PR #1408, opened 2019-08-06
INDY-2199: Endorsers must be specified within the transaction

ashcherbakov:
- Endorser field can have one of the following roles only: TRUSTEE, STEWARD or ENDORSER
- Endorser field must be explicitly present if transaction is trying to be endorsed with a multi-sig.


design/transaction_endorser.md
```diff
@@ -62,6 +63,12 @@ pub extern fn indy_append_request_endorser(command_handle: CommandHandle,
   - there must be `identifier`'s signature in `signatures`
   - `signature` must be absent
       
+#### Request dynamic validation
+In order to avoid endorsement without explicitly specifying an Endorser, the following changes to dynamic validation must be implemented: 
+- If there is `endorser` field, then it must have a known role (Trustee, Steward or Endorser).
```
skhoroshavin: It seems like according to new requirements it can have only Endorser role

indy_common/authorize/authorizer.py
```diff
@@ -192,3 +197,72 @@ def authorize(self,
                   auth_constraint: AuthConstraintForbidden,
                   auth_action: AbstractAuthAction):
         return False, str(auth_constraint)
+
+
+class EndorserAuthorizer(AbstractAuthorizer):
+    NO_NEED_FOR_ENDORSER_ROLES = {ENDORSER, TRUSTEE, STEWARD}
+    NO_NEED_FOR_ENDORSER_ROLES_STR = [ENDORSER_STRING, TRUSTEE_STRING, STEWARD_STRING]
```
skhoroshavin: Any reason why this is a list instead of set?

ashcherbakov: It's used in an error message only, and the order is important there while there is no checks for membership.

indy_common/test/auth/metadata/helper.py
```diff
@@ -39,61 +43,66 @@ def set_auth_constraint(validator, auth_constraint):
...
     action = AuthActionAdd(txn_type=req.operation[TYPE],
                            field='some_field',
                            value='new_value',
-                           is_owner=need_to_be_owner)
+                           is_owner=action.is_owner)
 
     return req, [action]
 
 
 def validate(auth_constraint,
```
skhoroshavin: I think adding docstrings to `validate`, `is_expected` and `build_req_and_action` could help a lot with tests readability

indy_common/test/auth/multi_sig/test_auth_multi_sig_for_1_owner.py
```diff
@@ -16,8 +16,11 @@ def test_claim_def_adding_success_1_owner(write_request_validation, req,
 
 
 def test_claim_def_adding_success_2_owner(write_request_validation, req,
-                                          identity_owners, key):
+                                          identity_owners, endorsers, key):
     req.signatures = {idr: "signature" for idr in identity_owners[:2]}
+    # Endorser must be present in case of multi-sig
```
skhoroshavin: Why? Before this PR this test showed, that signatures of two identity owners are enough for passing validation, so I suppose they had enough permissions. Now we have an endorser, and it is said in documentation that if original author of transaction have write access then endorser is not needed. Probably I'm missing something...

indy_node/test/nym_txn/test_create_did_without_endorser.py
```diff
@@ -101,27 +97,6 @@ def test_create_did_without_endorser_with_different_dest(looper, nym_txn_data, s
         sdk_get_and_check_replies(looper, [request_couple])
 
 
-def test_create_did_without_endorser_sig_count_2_one_on_ledger(looper, txnPoolNodeSet, nym_txn_data, sdk_pool_handle,
-                                                               sdk_wallet_trustee):
```
skhoroshavin: Any reason for removal of this test? I think I saw similar test, but I'm not sure, and it was among unit tests, not integration. So, dropping comment here just in case.

ashcherbakov: This test was just removed to another folder (as it requires endorsers now).

indy_common/test/auth/multi_sig/test_auth_multi_sig_for_1_owner.py
ashcherbakov: This is because of a simplistic approach (see PoA) to require an endorser in case of any muti-signature if the author has unprivileged role. The motivation was because endorsement can be done in different ways by different roles and depends on Auth Constraints (probably composite ones). 
But you are right, I think we can easily enhance it to not require Endorser if there is a multi-signature from unprivileged roles only.

ashcherbakov: Will do it in a next PR

PR #1417, opened 2019-08-22
INDY-2211: don't write OFF_LEDGER_SIGNATURE  if it's not explicitly set

Toktar:

### DIFF
--- a/indy_node/test/auth_rule/test_auth_rule_transaction.py
+++ b/indy_node/test/auth_rule/test_auth_rule_transaction.py
@@ -1,13 +1,17 @@
 import pytest
 import json
 
+from common.serializers.serialization import config_state_serializer
+from indy_common.state import config
+from indy_node.server.request_handlers.config_req_handlers.auth_rule.static_auth_rule_helper import StaticAuthRuleHelper
 from plenum.common.constants import TRUSTEE, STEWARD
 from plenum.common.exceptions import RequestRejectedException, \
     RequestNackedException
 
 from indy_common.authorize.auth_actions import ADD_PREFIX, EDIT_PREFIX
-from indy_common.authorize.auth_constraints import ROLE
-from indy_common.constants import AUTH_ACTION, OLD_VALUE
+from indy_common.authorize.auth_constraints import ROLE, CONSTRAINT_ID, ConstraintsEnum, SIG_COUNT, NEED_TO_BE_OWNER, \
+    METADATA, OFF_LEDGER_SIGNATURE
+from indy_common.constants import AUTH_ACTION, OLD_VALUE, NYM, ENDORSER, CONFIG_LEDGER_ID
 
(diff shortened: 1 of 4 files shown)


indy_node/test/helper.py
```diff
@@ -186,7 +186,7 @@ def sdk_send_and_check_auth_rule_request(
 
     # temp fix untill new sdk released
     req_json = json.loads(req_json)
-    if req_json[OPERATION][CONSTRAINT][CONSTRAINT_ID] == 'ROLE':
+    if req_json[OPERATION][CONSTRAINT][CONSTRAINT_ID] == 'ROLE' and OFF_LEDGER_SIGNATURE in constraint:
```
ashcherbakov: Since SDK supports OFF_LEDGER_SIGNATURE  now, do we still need this temp fix?

indy_node/test/auth_rule/test_constraint_field.py
```diff
@@ -131,3 +131,25 @@ def test_auth_constraint_from_dct_succesfull():
...
+
+    for k in auth_dct:
+        if k != unkwn:
+            assert hasattr(constr, k)
+    assert not hasattr(constr, unkwn)
+    assert constr.off_ledger_signature is None
+
+    del auth_dct[unkwn]
+    print(auth_dct)
+    print(constr.as_dict)
```
skhoroshavin: Why prints instead of logger?

indy_node/test/auth_rule/test_auth_rule_transaction.py
```diff
@@ -116,3 +120,33 @@ def test_reqnack_auth_rule_add_transaction_with_wrong_format(looper,
...
+    auth_type = NYM
+    field = ROLE
+    new_value = ENDORSER
+    constraint = {CONSTRAINT_ID: ConstraintsEnum.ROLE_CONSTRAINT_ID,
+                  ROLE: "*",
+                  SIG_COUNT: 1,
+                  NEED_TO_BE_OWNER: False,
+                  METADATA: {}}
+    if off_ledger_signature is not None:
+        constraint[OFF_LEDGER_SIGNATURE] = off_ledger_signature
```
ashcherbakov: Since SDK doesn't send OFF_LEDGER_SIGNATURE in request if it's sent to False, I think we need to expect it in `constraint` only if off_ledger_signature is True.

Toktar:
GitHub problems

indy_node/test/auth_rule/test_constraint_field.py
Toktar: Changes in https://github.com/hyperledger/indy-node/pull/1418

indy_node/test/helper.py
Toktar: Changes in https://github.com/hyperledger/indy-node/pull/1418

PR #1420, opened 2019-08-26
[INDY-2211] Add migration script

anikitinDSR:

### DIFF
--- a/data/migrations/deb/1_9_1_to_1_9_2.py
+++ b/data/migrations/deb/1_9_1_to_1_9_2.py
@@ -1,0 +1,77 @@
+import os
+import shutil
+import subprocess
+import sys
+
+from indy_common.config_helper import NodeConfigHelper
+from indy_common.config_util import getConfig
+from stp_core.common.log import getlogger
+
+logger = getlogger()
+
+ENV_FILE_PATH = "/etc/indy/indy.env"
+BUILDER_NET_NETWORK_NAME = "net3"
+
+
+def get_node_name():
+    node_name = None
+    node_name_key = 'NODE_NAME'
+
+    if os.path.exists(ENV_FILE_PATH):
+     with open(ENV_FILE_PATH, "r") as fenv:
+         for line in fenv.readlines():
+             if line.find(node_name_key) != -1:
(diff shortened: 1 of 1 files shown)


data/migrations/deb/1_9_1_to_1_9_2.py
```diff
@@ -0,0 +1,79 @@
...
+
+def is_node_stopped():
+    ret = subprocess.run(["systemctl", "status", "indy-node"], stdout=subprocess.PIPE)
+
+    if ret.returncode == 0:
+        return False
+    return True
+
+
+def migrate_all():
```
ashcherbakov: Where do we call the function?

data/migrations/deb/1_9_1_to_1_9_2.py
```diff
@@ -0,0 +1,79 @@
...
+
+    # Check, that node is stopped
```
ashcherbakov: Not sure that we need these steps, since this is part of migration logic, and the node will be stopped for sure.

data/migrations/deb/1_9_1_to_1_9_2.py
```diff
@@ -0,0 +1,79 @@
...
+    node_name = get_node_name()
+    if node_name is None:
+        return False
+
+    config = getConfig()
+
+    config_helper = NodeConfigHelper(node_name, config)
+
+    if BUILDER_NET_NETWORK_NAME != config.NETWORK_NAME:
+        logger.error("This script can be used only for {} network".format(BUILDER_NET_NETWORK_NAME))
```
ashcherbakov: This is not an error. We expect that this will do nothing on other Nets.

data/migrations/deb/1_9_1_to_1_9_2.py
```diff
@@ -0,0 +1,79 @@
...
+
+    try:
+        shutil.rmtree(path_to_config_state)
+    except shutil.Error as ex:
+        logger.error("""While removing directory: {}
+the next error was raised:
+{}""".format(path_to_config_state, ex))
+
+        return False
+
```
ashcherbakov: Should we also remove config's timestamp storage?

PR #1434, opened 2019-08-29
Release notes 1.9.2

VladimirWork:


VladimirWork:
Some versions will be changed with the next commit.

PR #1435, opened 2019-09-03
bump system tests

VladimirWork:
Signed-off-by: Vladimir Shishkin <vladimir.shishkin@dsr-corporation.com>

### DIFF
--- a/Jenkinsfile.nightly
+++ b/Jenkinsfile.nightly
@@ -43,11 +43,12 @@ try {
         pkgVersion = buildPkgVersion
         srcVersion = buildSrcVersion
         testSchema = [
-            ['test_ledger.py'],
-            ['test_vc.py'],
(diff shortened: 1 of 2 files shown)


Jenkinsfile.nightly
```diff
@@ -43,11 +43,12 @@ try {
         pkgVersion = buildPkgVersion
         srcVersion = buildSrcVersion
         testSchema = [
-            ['test_ledger.py'],
-            ['test_vc.py'],
+            ['test_ledger.py', 'test_state_proof'],
```
ashcherbakov: BTW maybe it makes sense to run `test_state_proof` on every build, not on nightly only? This looks like quite important functionality (maybe the most important for the end users).

VladimirWork: Done.

PR #1441, opened 2019-09-05
update playbooks

VladimirWork:
Signed-off-by: Vladimir Shishkin <vladimir.shishkin@dsr-corporation.com>

### DIFF
--- a/scripts/ansible/pool_get_logs.yml
+++ b/scripts/ansible/pool_get_logs.yml
@@ -1,6 +1,9 @@
 - hosts: pool
   gather_facts: no
 
+  vars:
+    pool_prefix: persistent_node
+
   tasks:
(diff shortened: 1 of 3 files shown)


scripts/ansible/pool_get_logs.yml
```diff
@@ -30,3 +33,27 @@
       src: /var/log/indy/sandbox/
       dest: ./logs
```
anikitinDSR: if it possible, please change to absolute path instead of relative

VladimirWork: Each scripts user forks indy-node in different place so probably it's better to keep relative paths for local machine.

VladimirWork:
test this please

VladimirWork:
test this please

PR #1507, opened 2019-11-15
[INDY-2292]: change state restoring logic

Toktar:


indy_node/server/request_handlers/config_req_handlers/auth_rule/abstract_auth_rule_handler.py
```diff
@@ -29,8 +33,24 @@ def _static_validation_for_rule(self, operation, identifier, req_id):
...
+    def _update_auth_constraint_for_1_9_1(self, auth_key: str, constraint):
+        constraint_dict = constraint.as_dict if constraint else {}
+        self._set_off_ledger_to_constraint(constraint_dict)
         self.state.set(AbstractAuthRuleHandler.make_state_path_for_auth_rule(auth_key),
-                       self.constraint_serializer.serialize(constraint))
+                       self.constraint_serializer.serializer.serialize(constraint_dict))
+
+    def _set_off_ledger_to_constraint(self, constraint_dict):
+        if constraint_dict[CONSTRAINT_ID] == ConstraintsEnum.ROLE_CONSTRAINT_ID:
+            constraint_dict[OFF_LEDGER_SIGNATURE] = None
```
ashcherbakov: Why do we set it to None here? I think it needs to be just explicitly set regardless of if this is True or False (now it's explicitly set if it's True only).

indy_node/test/catchup/test_state_recovering.py
```diff
@@ -0,0 +1,133 @@
...
+                            valid_upgrade)
+    send_node_upgrades(nodeSet, version1, looper, count=1)
+    send_auth_rule(looper,
+                   sdk_pool_handle,
+                   sdk_wallet_trustee,
+                   auth_keys[0])
+    send_node_upgrades(nodeSet, version1, looper)
+    for n in nodeSet:
+        handler = n.write_manager.request_handlers.get(AUTH_RULE)[0]
+        monkeypatch.setattr(handler, '_update_auth_constraint',
```
ashcherbakov: Why do we do monkey patching here? This doesn't check that we chose valid state update strategy, does it?

indy_node/server/request_handlers/config_req_handlers/auth_rule/abstract_auth_rule_handler.py
```diff
@@ -18,6 +20,8 @@ def __init__(self, database_manager: DatabaseManager,
         super().__init__(database_manager, txn_type, CONFIG_LEDGER_ID)
         self.write_req_validator = write_req_validator
         self.constraint_serializer = ConstraintsSerializer(domain_state_serializer)
+        self.config = getConfig()
+        self._update_state_by_versions = self._get_update_state_by_versions()
```
ashcherbakov: Although this version processing can be fine for the fix, I think it would be great to eventually do it in a cleaner way using strategy pattern more explicitly:
1) Create AuthRule Handler classes for the version 1.9.1 (the only thing that they do differently is `update_state` method).
2) Add a possibility to register req handlers for the given state
3) Register AuthRule1.9.1
4) Apply a strategy in WriteRegManager's `restore_state`:
- get a version for the txn
- if there is a specific req handler for this version registered - call it
- otherwise call a common (current) handler
5) In future: extend the versioning pattern to other or all WriteRegManager's methods

indy_node/test/catchup/test_state_recovering.py
Toktar: We patch ```update_state``` there to write AUTH_RULE txns in the old format for ordinary nodes.

Toktar:
test this please

indy_node/server/request_handlers/config_req_handlers/auth_rule/auth_rule_handler_1_9_1.py
```diff
@@ -0,0 +1,27 @@
...
+
+    def _update_auth_constraint(self, auth_key: str, constraint):
+        constraint_dict = constraint.as_dict if constraint else {}
+        self._set_off_ledger_to_constraint(constraint_dict)
+        self.state.set(AbstractAuthRuleHandler.make_state_path_for_auth_rule(auth_key),
+                       self.constraint_serializer.serializer.serialize(constraint_dict))
+
+    def _set_off_ledger_to_constraint(self, constraint_dict):
+        if constraint_dict[CONSTRAINT_ID] == ConstraintsEnum.ROLE_CONSTRAINT_ID:
+            constraint_dict[OFF_LEDGER_SIGNATURE] = False
```
ashcherbakov: Please do this only if `OFF_LEDGER_SIGNATURE` is not True

PR #1514, opened 2019-11-20
INDY-2292: add more tests for a state restoring

Toktar:

### DIFF
--- a/indy_node/test/request_handlers/test_auth_rule_1_9_1_handler.py
+++ b/indy_node/test/request_handlers/test_auth_rule_1_9_1_handler.py
@@ -1,0 +1,27 @@
+import pytest
+from indy_common.authorize.auth_constraints import OFF_LEDGER_SIGNATURE
+from indy_node.server.request_handlers.config_req_handlers.auth_rule.abstract_auth_rule_handler import \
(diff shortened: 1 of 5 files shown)


indy_node/test/request_handlers/test_auth_rule_1_9_1_handler.py
```diff
@@ -0,0 +1,27 @@
...
+    txn = reqToTxn(auth_rule_request)
+    payload = get_payload_data(txn)
+    constraint = StaticAuthRuleHelper.get_auth_constraint(payload)
+    auth_key = StaticAuthRuleHelper.get_auth_key(payload)
+    path = AbstractAuthRuleHandler.make_state_path_for_auth_rule(auth_key)
+
+    auth_rule_handler_1_9_1.update_state(txn, None, auth_rule_request)
+    expected_dict = constraint.as_dict
+    expected_dict[OFF_LEDGER_SIGNATURE] = False
+    assert auth_rule_handler_1_9_1.get_from_state(path) == expected_dict
```
ashcherbakov: A general comment: I think we need to explicitly check for expected state key, and maybe state value, to reduce risks of unintentianl change of a key.

PR #1527, opened 2019-12-03
[INDY-2303] Fix dependency list compiling for upgrade procedure

anikitinDSR:


indy_node/utils/node_control_utils.py
```diff
@@ -357,18 +357,25 @@ def get_deps_tree(cls, *package, depth=0):
...
+            # Make deps list unique
+            deps = list(set(deps + ret))
+            for d in deps:
+                if "=" in d:
+                    p, v = d.split("=")
+                else:
+                    p = d
+                    v = ''
+                if p in hold_list and \
+                        (p not in deps_map or deps_map[p] == ''):
```
ashcherbakov: Do we have unit tests where we have the same dependency in a tree with a version and without a version?

anikitinDSR: Yes, in test `test_deps_levels` we have pkg_a, which depends from `bb` and `cc` with versions, but package `bb` depends from package `cc` without version.

PR #1539, opened 2019-12-11
[INDY-2306]: Bump plenum version

donqui:


anikitinDSR:
PR with higher version of indy-plenum was merged

PR #1543, opened 2019-12-19
INDY-2302/2313: Update Transaction Author Agreement documentaion

skhoroshavin:

### DIFF
--- a/docs/source/requests.md
+++ b/docs/source/requests.md
@@ -17,6 +17,7 @@
     * [AUTH_RULES](#auth_rules)
     * [TRANSACTION_AUTHOR_AGREEMENT](#transaction_author_agreement)
     * [TRANSACTION_AUTHOR_AGREEMENT_AML](#transaction_author_agreement_AML)    
+    * [TRANSACTION_AUTHOR_AGREEMENT_DISABLE](#transaction_author_agreement_disable)
     * [SET_CONTEXT](#set_context)
 
 * [Read Requests](#read-requests)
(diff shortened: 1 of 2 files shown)


docs/source/requests.md
```diff
@@ -1723,30 +1724,42 @@ A client will receive NACK for
 Setting (enabling/disabling) a transaction author agreement for the pool.
 If transaction author agreement is set, then all write requests to Domain ledger (transactions) must include additional metadata pointing to the latest transaction author agreement's digest which is signed by the transaction author.
 
-If no transaction author agreement is set, or it's disabled, then no additional metadata is required.
+For any given version of transaction author agreement text and ratification date cannot be changed once set. Ratification date cannot be in future.
```
ashcherbakov: I would exchange these two paragraphs since the second one is related to the very first one (client's metadata in request).

docs/source/requests.md
```diff
@@ -1723,30 +1724,42 @@ A client will receive NACK for
...
 
-Transaction author agreement can be disabled by setting an agreement with an empty text.
+If no transaction author agreement is set, or there are no active transaction author agreements, then no additional metadata is required.
 
-Each transaction author agreement has a unique version.
+Individual transaction author agreements can be disabled by setting retirement date using same transaction. Retirement date can be in future, in this case deactivation of agreement won't happen immediately, it will be automatically deactivated at required date instead. It is also possible to change (or delete) existing retirement date of agreement if it didn't occur yet.
```
ashcherbakov: Currently we allow to check retirement date for any TAA regardless of the current retirement value, that is we can re-enforce an old TAA. So, we should either fix the implementation  or the documentation.

docs/source/requests.md
```diff
@@ -1723,30 +1724,42 @@ A client will receive NACK for
...
 
 - `version` (string):
 
     Unique version of the transaction author agreement
 
-
-- `text` (string):
+- `text` (string; optional):
 
     Transaction author agreement's text
```
ashcherbakov: Please mention that it's must-have for new TAAs, and optional for updating of existing TAAs (so for updates it should either not be specified at all or be equal to initial value)

docs/source/requests.md
```diff
@@ -1723,30 +1724,42 @@ A client will receive NACK for
...
 
+- `ratified` (integer; optional):
```
ashcherbakov: Please mention that it's must-have for new TAAs and can not be changed latter (so for updates it should not be specified at all, or must be equal to the initial value).
Also please mention what precision is expected here (actually any).

ashcherbakov: BTW it may make sense to check if we mention that TAA acceptance date in requests must be rounded to date only.

docs/source/requests.md
```diff
@@ -1723,30 +1724,42 @@ A client will receive NACK for
...
+
+    Timestamp of Transaction Author Agreement ratification date
+
+- `retired` (integer; optional):
```
ashcherbakov: Please mention that it can not be set for new TAA. Please mention that absense or None value in TAA updates means cancelling retirement or re-enforcing the TAA after retirement. 
Also please mention what precision is expected here (actually any).

ashcherbakov: Also please mention that we can not set retirement date for a TAA after disable all action if there is no latest TAA defined yet (and after disable all action there is no latest TAA).

docs/source/requests.md
```diff
@@ -1723,30 +1724,42 @@ A client will receive NACK for
...
+
+    Timestamp of Transaction Author Agreement retirement date.
+
 *Request Example*:
```
ashcherbakov: I think it makes sense to provide two examples here: one for adding and one for update (retirement)

PR #1562, opened 2020-01-09
[INDY-2314] add size limitation for revocation's field.

anikitinDSR:


indy_common/types.py
```diff
@@ -155,26 +155,26 @@ class RevocDefValueField(MessageValidator):
...
 
 
 class ClientRevocDefSubmitField(MessageValidator):
     schema = (
         (TXN_TYPE, ConstantField(REVOC_REG_DEF)),
-        (ID, NonEmptyStringField()),
-        (REVOC_TYPE, NonEmptyStringField()),
-        (TAG, NonEmptyStringField()),
-        (CRED_DEF_ID, NonEmptyStringField()),
+        (ID, LimitedLengthStringField()),
```
ashcherbakov: I think it can be more than 256 chars

indy_common/types.py
```diff
@@ -155,26 +155,26 @@ class RevocDefValueField(MessageValidator):
...
+        (REVOC_TYPE, LimitedLengthStringField()),
+        (TAG, LimitedLengthStringField(max_length=TAG_LIMIT_SIZE)),
+        (CRED_DEF_ID, LimitedLengthStringField()),
```
ashcherbakov: I think it can be more than 256 chars

indy_common/types.py
```diff
@@ -155,26 +155,26 @@ class RevocDefValueField(MessageValidator):
...
         (VALUE, RevocDefValueField())
     )
 
 
 class RevocRegEntryValueField(MessageValidator):
     schema = (
-        (PREV_ACCUM, NonEmptyStringField(optional=True)),
-        (ACCUM, NonEmptyStringField()),
+        (PREV_ACCUM, LimitedLengthStringField(optional=True)),
+        (ACCUM, LimitedLengthStringField()),
```
ashcherbakov: I think it can be more than 256 chars

indy_common/types.py
```diff
@@ -183,8 +183,8 @@ class RevocRegEntryValueField(MessageValidator):
 class ClientRevocRegEntrySubmitField(MessageValidator):
     schema = (
         (TXN_TYPE, ConstantField(REVOC_REG_ENTRY)),
-        (REVOC_REG_DEF_ID, NonEmptyStringField()),
-        (REVOC_TYPE, NonEmptyStringField()),
+        (REVOC_REG_DEF_ID, LimitedLengthStringField()),
```
ashcherbakov: I think it can be more than 256 chars

indy_node/test/anon_creds/test_revoc_def_static_validation.py
```diff
@@ -0,0 +1,62 @@
...
+from plenum.test.helper import sdk_sign_request_from_dict, sdk_send_signed_requests, sdk_get_replies
+
+
+
+@pytest.fixture(scope="module", params=['lt', 'eq', 'gt'])
+def _lt_eq_gt(request):
+    return request.param
+
+
+@pytest.fixture(scope="module", params=[ID, REVOC_TYPE, TAG, CRED_DEF_ID, TAILS_HASH, TAILS_LOCATION])
```
ashcherbakov: Should we still have `ID` and `CRED_DEF_ID` here?

indy_common/test/types/test_revoc_def_submit_schema.py
```diff
@@ -9,15 +10,15 @@
     (ISSUANCE_TYPE, ChooseField),
     (MAX_CRED_NUM, IntegerField),
     (PUBLIC_KEYS, AnyMapField),
-    (TAILS_HASH, NonEmptyStringField),
-    (TAILS_LOCATION, NonEmptyStringField),
+    (TAILS_HASH, LimitedLengthStringField),
```
ashcherbakov: Are we sure that tails hash has not more than 256 bits? What if it is SHA-512?

indy_node/test/anon_creds/test_revoc_def_static_validation.py
```diff
@@ -0,0 +1,57 @@
...
+
+@pytest.fixture(scope="module")
+def revoc_def_req(looper,
+                  sdk_wallet_steward,
+                  build_revoc_def_by_default,
+                  _res_field_size):
+    _expected, _field, _size = _res_field_size
+    _req = copy.deepcopy(build_revoc_def_by_default)
+
+    if _field in (TAILS_HASH, TAILS_LOCATION):
```
ashcherbakov: Do we need this condition? It uses `randomString` in both branches.

anikitinDSR: it needs because we keep `TAILS_HASH` and `TAILS_LOCATION` not in `operation` section directly but in `value`.

indy_common/test/types/test_revoc_def_submit_schema.py
anikitinDSR: Should we drop limitation for this field or add some specific limit? Maybe max(current possible hashes) * 10?

PR #1604, opened 2020-06-02
nym dynamic validation check transaction w/out verkey or role

brentzundel:
modifies nym transaction dynamic validation to check for `role` and `verkey` properties in the operation, fails if both are missing.

### DIFF
--- a/indy_node/server/request_handlers/domain_req_handlers/nym_handler.py
+++ b/indy_node/server/request_handlers/domain_req_handlers/nym_handler.py
@@ -109,10 +109,10 @@ class NymHandler(PNymHandler):
         is_owner = origin == owner
 
         updateKeys = [ROLE, VERKEY]
-        updateKeysInOperation = is_owner
+        updateKeysInOperationOrOwner = is_owner
         for key in updateKeys:
             if key in operation:
-                updateKeysInOperation = True
+                updateKeysInOperationOrOwner = True
                 newVal = operation[key]
                 oldVal = nym_data.get(key)
                 self.write_req_validator.validate(request,
@@ -121,7 +121,7 @@ class NymHandler(PNymHandler):
                                                                   old_value=oldVal,
                                                                   new_value=newVal,
                                                                   is_owner=is_owner)])
-        if not updateKeysInOperation:
+        if not updateKeysInOperationOrOwner:
             raise InvalidClientRequest(request.identifier, request.reqId)
 
     def _decode_state_value(self, encoded):
--- a/indy_node/server/request_handlers/domain_req_handlers/nym_handler.py
+++ b/indy_node/server/request_handlers/domain_req_handlers/nym_handler.py
@@ -109,8 +109,10 @@ class NymHandler(PNymHandler):
         is_owner = origin == owner
 
         updateKeys = [ROLE, VERKEY]
+        updateKeysInOperation = is_owner
         for key in updateKeys:
             if key in operation:
+                updateKeysInOperation = True
                 newVal = operation[key]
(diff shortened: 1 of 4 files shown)


brentzundel:
@dhh1128 the only thing this PR changes is to prevent a nym transaction from succeeding if:
1. the DID is already on the ledger (the nym transaction is an update to an existing DID)
1. the submitting DID is not the same as the DID being updated
1. the nym transaction does not update the `role` nor the `verkey` property of the existing DID (those cases are handled by the auth rules)

None of these restrictions on a nym transaction should affect tokens or token transactions in any way, as far as I understand them.

dhh1128:
>None of these restrictions on a nym transaction should affect tokens or token transactions in any way, as far as I understand them.

Okay. Cross that off my list of concerns.

I'm still hesitant about this:

>the submitting DID is not the same as the DID being updated

Is it supposed to be possible for a TRUSTEE to update a different DID? (I'm thinking no, in which case the restriction as you have it is perfect. If yes, then we've just locked them out.)

brentzundel:
> Is it supposed to be possible for a TRUSTEE to update a different DID? (I'm thinking no, in which case the restriction as you have it is perfect. If yes, then we've just locked them out.)

It is still possible for a TRUSTEE to update a different DID's `role` or `verkey` (as long as the auth rules allow this), even with this change.

Toktar:
(CI) Test this please

indy_node/test/nym_txn/test_nym_auth_rules.py
```diff
@@ -334,4 +334,8 @@ def test_nym_edit(
     if editor.verkey is None:  # skip that as well since it doesn't make sense
         return
 
+    if not ROLE in edit_op:  # skip if the update operation doesn't changes neither role nor verkey
```
brentzundel, line 337:
```suggestion
    if not ROLE in edit_op:  # skip if the update operation changes neither role nor verkey
```

brentzundel:
@Toktar @skhoroshavin can this get review please? It is a bug fix and would be good to get merged.

indy_node/server/request_handlers/domain_req_handlers/nym_handler.py
```diff
@@ -109,8 +109,10 @@ def _validate_existing_nym(self, request, operation, nym_data):
         is_owner = origin == owner
 
         updateKeys = [ROLE, VERKEY]
+        updateKeysInOperation = is_owner
```
brentzundel, line 112:
```suggestion
        updateKeysInOperationOrOwner = is_owner
```

indy_node/server/request_handlers/domain_req_handlers/nym_handler.py
```diff
@@ -109,8 +109,10 @@ def _validate_existing_nym(self, request, operation, nym_data):
...
         for key in updateKeys:
             if key in operation:
+                updateKeysInOperation = True
```
brentzundel, line 115:
```suggestion
                updateKeysInOperationOrOwner = True
```

indy_node/server/request_handlers/domain_req_handlers/nym_handler.py
```diff
@@ -119,6 +121,8 @@ def _validate_existing_nym(self, request, operation, nym_data):
                                                                   old_value=oldVal,
                                                                   new_value=newVal,
                                                                   is_owner=is_owner)])
+        if not updateKeysInOperation:
```
brentzundel, line 124:
```suggestion
        if not updateKeysInOperationOrOwner:
```

PR #1622, opened 2020-10-03
add build file for github actions

m00sey:
This PR is superseded by #1645 

~Jobs for indy-common/node, lint and `@pytest.mark` check.~

~In order to parallelize the build, I created a build matrix for each of the "test" packages.~
~Each test is annotated with `@pytest.mark<package>`~

~The pytest_checks job returns an array of all found unique decorators and uses that as the build matrix.~

~The build will fail if any `test_` method is missing a mark decorator.~

~This takes the build time from >45 minutes to 12 minutes.~

Signed-off-by: Kevin Griffin <griffin.kev@gmail.com>


Toktar:
You have done an amazing job! Thank you.
But just in case, I want to warn that Plenum https://github.com/hyperledger/indy-plenum is a large part of Indy Node, they are inextricably linked. And it might be more convenient to start with it.

setup.py
```diff
@@ -28,7 +28,7 @@
 BASE_DIR = os.path.join(os.path.expanduser("~"), ".indy")
 
 tests_require = ['attrs==19.1.0', 'pytest==3.3.1', 'pytest-xdist==1.22.1', 'pytest-forked==0.2',
-                 'python3-indy==1.13.0-dev-1420', 'pytest-asyncio==0.8.0']
+                 'python3-indy==1.15.0', 'pytest-asyncio==0.8.0']
```
WadeBarnes, line 31: @jovfer, @askolesov, @Toktar, What is your opinion on this?  Do you anticipate any unforeseen issues here.

.github/workflows/build.yaml
```diff
@@ -0,0 +1,88 @@
+name: indy-node-build
+on: [ push, pull_request ]
+
+jobs:
+  decorator-checks:
+    name: Run decorator checks
+    needs: lint
+    runs-on: ubuntu-18.04
+    #TODO: move this to hyperledger?
+    container: m00sey/indy-node-build:latest
```
WadeBarnes, line 10: @m00sey, This references will need to be updated.

WadeBarnes (changes requested):
@m00sey, Looking good!  Just a few references to update that I can see.

I'd also like to see some documentation indicating which Jenkins pipeline flow(s) this replaces, to make it easier to:
1. Identify the Jenkins workflows being replaced.
2. Review and compare the Jenkins vs GHA flows to verify everything is covered in the new workflows.
3. Decommission the Jenkins workflows when we're satisfied we have equivalent or better GHA flows.

WadeBarnes:
(ci) test this please

setup.py
askolesov: I think it's ok if tests are passing

WadeBarnes: They are.

WadeBarnes:
@m00sey, I think some documentation around how to build and manage the `indy-node-build` container used by the actions would be useful too.

.github/workflows/build.yaml
m00sey: By updated do you mean pushing the image to hyperledger?

WadeBarnes: Ultimately, yes.  However we should discuss how to manage the image since it contains dependencies that will need to be updated occasionally.

setup.py
Toktar: @WadeBarnes, @m00sey I really like this idea. But if we change the dependency here, I suggest changing them in `ci/uduntu.dockerfile` to preservation of consistency even if this file will be deprecated soon.
And it's a bit strange if indy-node and indy-plenum will depend on different libindy packages. I guess it's absolutely unimportant because of it's used just for a test environment, but still.

.github/workflows/build.yaml
```diff
@@ -0,0 +1,158 @@
...
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v2
+
+      - name: Install dependencies
+        run: pip install .[tests]
+        continue-on-error: true
+
+      - name: Run Indy Node ${{ matrix.module }} tests
+        run: python3 -m pytest -l -m ${{ matrix.module }} -vv --junitxml=test-result-node-${{ matrix.module }}.xml indy_node
```
Toktar, line 136: The implementation of parallelizing tests looks pretty cool.
We already have `runner.py` for parallelizing tests. It would be great to replace it with something better, but maybe we could do it in the different PRs with adding a GitHub actions to make changes gradually?

m00sey: I'm not sure how runner.py would parallelize in the same way, given this issues a run command for each permutation in the matrix, runner.py would spawn a single runner.

indy_node/test/auth_rule/auth_framework/test_auth_rule_using.py
```diff
@@ -191,19 +159,157 @@ def env(self,
...
+    # def edit_role_actions_tests(self, request, env):
+    #     action_id = request.param
+    #     test_cls = self.edit_role_actions_test[action_id]
+    #     test = test_cls(env, action_id)
+    #     return action_id, test
+
+    @pytest.mark.auth_framework
+    def test_auth_rule_add_new_trustee(self, env):
+        self.runner(AddNewTrusteeTest(env, auth_map.add_new_trustee.get_action_id()))
+
```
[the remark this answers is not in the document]
m00sey: So the problem with this test was that the single test, output over 30,000 lines of logs, and is being pushed to the surefire plugin as a single string. 

This caused the surefire plugin to break.

The suite has been run many times, without an ordering failure.

However I understand your concern, if we could keep the separate test invocations, but, use pytest.mark to keep order, would that satisfy your concern?

.github/workflows/build.yaml
Toktar: It doesn't parallelize in the same way, but runner.py  has `--test-only-slice` to run only some part of tests dividing them on directories. 
Example of using runner.py  from indy-plenum CI: 
`RUSTPYTHONASYNCIODEBUG=0 python runner.py --pytest python -m pytest -l -vv --dir plenum --output test-result-plenum-1.jenkinsubuntu01.txt --test-only-slice 1/3` 
where **1/3** is a part which we want to launch.

m00sey: That particular runner and usage is already a part of the plenum GHA PR.

m00sey: I will look into the usage of slice for node.

indy_node/test/auth_rule/auth_framework/test_auth_rule_using.py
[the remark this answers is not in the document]
Toktar: Got it. Sure, with a same order it should permanently work.

Toktar:
[for CI] Test this please

WadeBarnes:
Closed in favor of https://github.com/hyperledger/indy-node/pull/1645

PR #1633, opened 2020-11-30
Implemented RichSchema feature flag

askolesov:


Toktar:
(ci) test this please

indy_common/config.py
```diff
@@ -113,3 +113,5 @@
 
 CONTEXT_SIZE_LIMIT = 131072
 JSON_LD_LIMIT = CONTEXT_SIZE_LIMIT
+
+enableRichSchemas = False
```
Toktar, line 117: Suggest changing `enableRichSchemas` to `ENABLE_RICH_SCHEMAS` because it's some kind of constants. But in fact, looks like we don't have rules for config naming. So it's just my point of view and not necessary to change.

Toktar (approved):

indy_node/server/request_handlers/read_req_handlers/get_rich_schema_object_by_id_handler.py
```diff
@@ -12,6 +14,10 @@ def __init__(self, database_manager: DatabaseManager):
 
     def get_result(self, request: Request):
         self._validate_request_type(request)
+
+        if not getConfig().enableRichSchemas:
+            raise InvalidClientRequest(request.identifier, request.reqId, "RicheSchemas feature is disabled")
```
Toktar, line 19: I think there is a small typo here `RichESchemas`
We can make the message more detailed. Something like "RichSchema transaction is disabled" and "GetRichSchema query is disabled". But the current variant is ok.

Toktar:
(ci) test this please

Toktar:
(ci) test this please

Toktar (approved):

indy_node/test/helper.py
```diff
@@ -324,3 +324,24 @@ def createCryptonym(seed=None):
...
+        setattr(self._obj, self._attr, self._new_value)
+        return self
+
+    def __exit__(self, exc_type, exc_value, exc_traceback):
+        setattr(self._obj, self._attr, self._old_value)
+
+
+# Returns context with RichSchemas feature enabled.
+def rich_schemas_enabled_scope(tconf):
+    return ModifiedAttributeScope(tconf, "ENABLE_RICH_SCHEMAS", True)
```
skhoroshavin, line 347: Instead of this you can create a short context manager, like here: https://github.com/hyperledger/indy-plenum/blob/master/plenum/test/helper.py#L1171

skhoroshavin (commented):
This looks great in general, but needs a little more improvements. Please see my comments, they should be really easy to fix.

indy_node/test/helper.py
askolesov: Such a cool thing! Thanks!

skhoroshavin (approved):

skhoroshavin:
(ci) ok to test

PR #1665, opened 2021-03-01
Merge Stable to Master

Toktar:


.github/workflows/build/Dockerfile
```diff
@@ -4,7 +4,7 @@ LABEL maintainer="Hyperledger <hyperledger-indy@lists.hyperledger.org>"
 RUN apt-get update -y && apt-get install -y \
     python3-nacl \
     libindy-crypto=0.4.5 \
-    libindy=1.13.0~1420 \
+    libindy=1.15.0~1618-xenial \
```
WadeBarnes, line 7: This change already exists in `master`.

Jenkinsfile.cd
```diff
@@ -93,7 +93,7 @@ def systemTests = { component, releaseVersion ->
             ['TestConsensusSuite.py'],
             ['test_state_proof.py']
         ]
-        testVersion = 'v0.8.41'
+        testVersion = 'v0.8.43'
```
WadeBarnes, line 96: This change already exists in `master`.

Jenkinsfile.ci
```diff
@@ -73,6 +73,7 @@ def withTestEnv(body) {
 
         buildDocker("hyperledger/indy-node-ci", "ci/ubuntu.dockerfile ci").inside {
             echo 'Test: Install dependencies'
+            sh "pip install pip==10.0.0"
```
WadeBarnes, line 76: This change already exists in `master`.

ci/ubuntu.dockerfile
```diff
@@ -8,7 +8,7 @@ ARG venv=venv
 RUN apt-get update -y && apt-get install -y \
     python3-nacl \
     ursa=0.3.2-2 \
-    libindy=1.13.0~1420 \
+    libindy=1.15.0~1618-xenial \
```
WadeBarnes, line 11: This change already exists in `master`.

WadeBarnes (changes requested):
There appear to be a number of identified changes that already exist in `master`, so I'm confused as to why they are appearing.

The PR also appears to be introducing a number of new files that don't (didn't) exist in `master` before.  Mostly in the `pool_automation` folder(s).  Is this a concern?

Toktar:
@WadeBarnes I agree, this is strange. But we have repeatedly caught this problem in the GitHub interface before because of rebases. The main thing is that the final version coincides with the one in the master. I checked it too.

Toktar:
@WadeBarnes Hopefully I've fixed the difference displaying issue.

scripts/indy
```diff
@@ -0,0 +1,84 @@
+#! /usr/bin/env python3
```
WadeBarnes, line 1: This file was deleted from `master` a long time ago; https://github.com/hyperledger/indy-node/commit/445452a4410f961c480a7344493932cc8090a221

I don't think it should be re-introduced.

WadeBarnes (changes requested):
The `pool_automation/roles/pool_install` and `pool_automation/roles/node_install` files and folder were deleted from `master` ( https://github.com/hyperledger/indy-node/commit/74d226d7937e7f50a834a15fb84117be45bcade8), I don't think they should be reintroduced.  It's looking like most or all of the "new" files under `pool_automation` are files that were previously deleted.

I'm also a little concerned about the changes to the files in the `data/migrations/deb` folder.  It appears as though they are being reverted (renamed and updated) to a previous state.  I think we should have someone more familiar with the process have a closer look.

Toktar:
- pool_automation - @skhoroshavin said it's okay to push these files to the Master.
- scripts/indy - it's the script from the old CLI. I'll remove it
- data/migrations/deb - I consulted with @spivachuk who added this renaming. So the master contains scripts for the master versions and the stable for the stable versions. I think it will be better to return master names, but it may be discussed.

Finally, we have 2 options:
- remove unnecessary commits from this PR
+clean history
-difference with stable (if we create a new PR Stable -> Master we will have a difference again)
- add a new commit with removing unnecessary files
+no difference between branches(I hope, I'm not sure), understandable history
-lots of modifies with files in a branch history (they are already strange because they were not made by commits from the links)

@WadeBarnes  What do you think?

Toktar:
The first option was turned out to be too expensive and difficult(impossible) to implement (that was mentioned on Indy Contributors call). 
PR is ready for review.

WadeBarnes:
@Toktar, I trying to wrap my head around the differences between the `data/migrations` in `master` and `stable` and the best way to manage the differences.  The scripts have clearly diverged over time.  You've indicated the ones in `master` are for the `master` versions and the ones in `stable` are the ones for the `stable` versions.

Can you provide me with (or point me at):
- some history on the differences between the `master` and `stable` versions.
- some history on why the `master` and `stable` versions needed different migration scripts.

If I'm reading the code correctly, migrations are only run when needed based on the current version installed on a node and the intended version to which the node will be upgraded.  Therefore with the current version of the code, none of the migration scripts would ever be run, since the newest migration script was for `1_9_1_to_1_9_2`.  Is this correct?

WadeBarnes:
There are some differences between this PR and the more direct one, https://github.com/hyperledger/indy-node/pull/1646, that I want to look into further, specifically related to differences in the following files that do not appear in this PR:
- indy_node/server/request_handlers/domain_req_handlers/nym_handler.py 
- indy_node/test/anon_creds/test_revoc_def_static_validation.py 
- indy_node/test/nym_txn/test_nym_auth_rules.py
- indy_node/test/nym_txn/test_nym_no_role.py 
- indy_node/test/request_handlers/test_nym_handler.py

Toktar:
> There are some differences between this PR and the more direct one, #1646, that I want to look into further, specifically related to differences in the following files that do not appear in this PR:
> 
> indy_node/server/request_handlers/domain_req_handlers/nym_handler.py
> indy_node/test/anon_creds/test_revoc_def_static_validation.py
> indy_node/test/nym_txn/test_nym_auth_rules.py
> indy_node/test/nym_txn/test_nym_no_role.py
> indy_node/test/request_handlers/test_nym_handler.py

These changes already in Master:
https://github.com/hyperledger/indy-node/blob/04ca6647200992052289051788b686034a7a2359/indy_node/server/request_handlers/domain_req_handlers/nym_handler.py#L112
They were added by this PR https://github.com/hyperledger/indy-node/pull/1604
I think in #1646 we see incorrect differences (maybe after rebase). I closed #1646  because work on it was stopped.

Toktar:
@WadeBarnes About versioning and migration scripts. 
As far as I remember, migration scripts are launch in case of upgrading pool from the first version to the second. I mean that for `1_9_1_to_1_9_2` migration script it's an upgrade from `1_9_1` to `1_9_2`

>  Therefore with the current version of the code, none of the migration scripts would ever be run, since the newest migration script was for 1_9_1_to_1_9_2. Is this correct?

Yes, this is correct because all known pools have the latest version of Indy Node. But we need to leave this code for remembering breaking changes and theoretically for some tests.

The same migration scripts have different names(versions) because for Master and for Stable brunches we have different channels in the repo with deb packages. So, for one breaking change we need a migration script with are triggered for some pair of Master package versions and for another pair of Stable package versions

Did it help? Can I answer some other questions?

WadeBarnes:
We are not planning any additional 16.04 releases.

PR #1702, opened 2021-10-26
Ubuntu16:04: Indy-Test-Automation workflow

udosson:
This PR provides a GHA workflow to run the node system tests from https://github.com/hyperledger/indy-test-automation
The workflow covers all tests from the following pipelines
- The CD Pipeline runs these set of tests; https://github.com/hyperledger/indy-node/blob/master/Jenkinsfile.cd#L87-L100
- The nightly build runs these set of tests; https://github.com/hyperledger/indy-node/blob/master/Jenkinsfile.nightly#L41-L72

- [ ] Test with Debian packages from Artifactory 
- [ ] Decide about the trigger of the pipeline

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -54,7 +54,7 @@ setup(
     data_files=[(
         (BASE_DIR, ['data/nssm_original.exe'])
     )],
-    install_requires=['indy-plenum==1.13.0.dev143',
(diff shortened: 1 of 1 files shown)


udosson:
Will be created from different branch

PR #1711, opened 2021-11-09
prepares indy-plenum package version of debian version depedency

udosson:
This change is needed to replace the PyPI version of `plenum` with the version of the Debian artifact.
Otherwise, we run into this error:
`indy-node : Depends: indy-plenum (= 1.13.0.dev169) but 1.13.0~dev169 is to be installed`

Signed-off-by: udosson <r.klemens@yahoo.de>


WadeBarnes:
(ci) test this please

WadeBarnes (approved):

PR #1716, opened 2021-11-26
Fixes wrong max slices in GHA

pSchlarb:
Signed-off-by: pSchlarb <p.schlarb@esatus.com>

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -199,6 +199,8 @@ jobs:
     # - https://github.com/ScaCap/action-surefire-report/issues/17
     env:
       NODE_OPTIONS: '--max_old_space_size=4096'
+      #SLICE_TOTAL_SLICES needs to match the total number of slices in the matrix strategy.
+      SLICE_TOTAL_SLICES: 11
     container:
       image: ghcr.io/${{ needs.workflow-setup.outputs.GITHUB_REPOSITORY_NAME }}/node-build:ubuntu-16-04
     strategy:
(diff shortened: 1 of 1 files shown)


udosson:
@pSchlarb thanks for the PR. Could you please provide some context and justify the proposed changes?

pSchlarb:
Yes, shimizut in the Rocketchat mentioned that the tests run by the GHA are only executing 11 of 22 slices per module. 
It seems that the variable used `strategy.job-total` is the sum of all slices across all modules.
So to correctly test all slices with the runner these changes are neccessary.

WadeBarnes:
I don't really like the idea of hard coding the number slices.  The original intent was to be able to update this section and have the test runs automatically adjust accordingly.  Obviously `strategy.job-total` is the number of modules * the number of slices.  We want it to just be the number of slices.  The question is how do we get that value without hard coding it.

```
      matrix:
        module: [indy_node, indy_common]
        slice: [1, 2, 3, 4 ,5, 6, 7, 8, 9, 10, 11]
```

WadeBarnes:
It does not seem to be a way to get the count of items included in the elements of a matrix.

What if we added a `slice_count` environment variable and had it set to the total number of slices, and documented it has to match the number of elements contained in `strategy.matrix.slice`.  It's still a bit of a hack, but better then having to change the value in multiple places is/when the number of elements contained in `strategy.matrix.slice` is changed.

pSchlarb:
I am fine with that but i would call it `SLICE_TOTAL_SLICES`.

.github/workflows/build.yaml
```diff
@@ -199,6 +199,8 @@ jobs:
     # - https://github.com/ScaCap/action-surefire-report/issues/17
     env:
       NODE_OPTIONS: '--max_old_space_size=4096'
+      # Total Number of Slices
+      SLICE_TOTAL_SLICES: 11
```
WadeBarnes, line 203: Please add a note to indicate the value needs to match the total number if items listed in the `strategy.matrix.slice` element.

WadeBarnes (changes requested):

WadeBarnes:
(ci) test this please

WadeBarnes (approved):

PR #1718, opened 2021-11-30
VSCode Devcontainer

pSchlarb:
Addition of VSCode Devcontainer files.
Signed-off-by: Philipp Schlarb <p.schlarb@esatus.com>

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -54,7 +54,7 @@ RUN apt-get update -y && apt-get install -y \
     make \
     # Indy Node and Plenum
     libssl1.0.0 \
-    ursa=0.3.2-1 \
+    ursa=0.3.2-2 \
     # Indy SDK
     libindy=1.15.0~1625-bionic \
     # Need to move libursa.so to parent dir
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -2,7 +2,7 @@
 
 # [Choice] Ubuntu version (use hirsuite or bionic on local arm64/Apple Silicon): hirsute, focal, bionic
 ARG VARIANT="focal"
-FROM mcr.microsoft.com/vscode/devcontainers/base:0-${VARIANT}
+FROM mcr.microsoft.com/vscode/devcontainers/base:${VARIANT}
 
 RUN apt-get update -y && apt-get install -y \
     # common stuff
@@ -54,7 +54,7 @@ RUN apt-get update -y && apt-get install -y \
     make \
     # Indy Node and Plenum
     libssl1.0.0 \
-    ursa=0.3.2-2 \
+    ursa=0.3.2-1 \
     # Indy SDK
     libindy=1.15.0~1625-bionic \
     # Need to move libursa.so to parent dir
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,76 @@
+# See here for image contents: https://github.com/microsoft/vscode-dev-containers/tree/v0.205.2/containers/ubuntu/.devcontainer/base.Dockerfile
+
+# [Choice] Ubuntu version (use hirsuite or bionic on local arm64/Apple Silicon): hirsute, focal, bionic
+ARG VARIANT="focal"
+FROM mcr.microsoft.com/vscode/devcontainers/base:0-${VARIANT}
+
+RUN apt-get update -y && apt-get install -y \
+    # common stuff
+    git \
+    wget \
+    gnupg \
+    apt-transport-https \
+    ca-certificates \
+    apt-utils \
+    curl \
+    jq
+
+# ========================================================================================================
+# Update repository signing keys
+# --------------------------------------------------------------------------------------------------------
+# Hyperledger
+RUN apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 9692C00E657DDE61 && \
+    # Sovrin
+    apt-key adv --keyserver keyserver.ubuntu.com --recv-keys CE7709D068DB5E88
+# ========================================================================================================
+
+# Plenum
+#  - https://github.com/hyperledger/indy-plenum/issues/1546
+#  - Needed to pick up rocksdb=5.8.8
+RUN echo "deb  https://hyperledger.jfrog.io/artifactory/indy focal dev"  >> /etc/apt/sources.list && \
+    echo "deb http://security.ubuntu.com/ubuntu bionic-security main"  >> /etc/apt/sources.list && \
+    echo "deb https://repo.sovrin.org/deb bionic master" >> /etc/apt/sources.list && \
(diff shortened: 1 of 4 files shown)


.devcontainer/Dockerfile
```diff
@@ -0,0 +1,76 @@
...
+    echo "deb http://security.ubuntu.com/ubuntu bionic-security main"  >> /etc/apt/sources.list && \
+    echo "deb https://repo.sovrin.org/deb bionic master" >> /etc/apt/sources.list && \
+    echo "deb https://repo.sovrin.org/sdk/deb bionic master" >> /etc/apt/sources.list
+
+RUN apt-get update -y && apt-get install -y \
+    # Python
+    python3-pip \
+    python3-nacl \
+    # rocksdb python wrapper
+    rocksdb=5.8.8 \
```
burdettadam, line 40: Could we use the latest version of rocksdb? v6.25.3?

burdettadam: I found the comment from line 28, the issue with plenum. Thanks. I understand the reason for version 5.8.8 now.

burdettadam (dismissed):

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,76 @@
...
+    zstd \
+    # fpm
+    ruby \
+    ruby-dev \
+    rubygems \
+    gcc \
+    make \
+    # Indy Node and Plenum
+    libssl1.0.0 \
+    ursa=0.3.2-1 \
```
WadeBarnes, line 57:
```suggestion
    ursa=0.3.2-2 \
```
The `ursa` version was updated in the 16.04 docker file (used for the GHA workflows) but not the 20.04 one yet.  We should update here and in the Dockerfile for the 20.0

WadeBarnes (changes requested):
One small recommended change.  Other than that I'm thinking we should have the pytest settings pre-configured so the development environment is more complete from the start.  Thoughts?

pSchlarb:
> Other than that I'm thinking we should have the pytest settings pre-configured so the development environment is more complete from the start.

@WadeBarnes Do you mean providing the workspace settings.json for VScode ? 
Its a good idea and already thought about it, but i am not quite sure.
If yes should it only be minimal settings for the tests or more customized?
Could it be problematic providing settings  that don't work well with other settings? 
To be honest i haven't that much experience with VSCode.

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,76 @@
...
+    zstd \
+    # fpm
+    ruby \
+    ruby-dev \
+    rubygems \
+    gcc \
+    make \
+    # Indy Node and Plenum
+    libssl1.0.0 \
+    ursa=0.3.2-2 \
```
udosson, line 57: Is `ursa=0.3.2-2` available for bionic or focal?
In the repositories I just find `ursa=0.3.2-1` (https://repo.sovrin.org/deb/pool/bionic/master/u/ursa/) and `ursa=0.1.1` (https://hyperledger.jfrog.io/ui/native/indy/pool/focal/dev/u/ursa)

pSchlarb: I took everything from the Dockerfile for the GHA. If thats problematic, we also have to change these.

WadeBarnes:
> > Other than that I'm thinking we should have the pytest settings pre-configured so the development environment is more complete from the start.
> 
> @WadeBarnes Do you mean providing the workspace settings.json for VScode ? Its a good idea and already thought about it, but i am not quite sure. If yes should it only be minimal settings for the tests or more customized? Could it be problematic providing settings that don't work well with other settings? To be honest i haven't that much experience with VSCode.

Yes, I was thinking some minimum settings for the tests, at least to start.  It shouldn't be an issue since they would be meant to run in the dev container which is where we get the consistency for the development environment.

.devcontainer/Dockerfile
WadeBarnes: Perhaps we need to keep it at `ursa=0.3.2-1` until we can get a version of `ursa=0.3.2-2` built for focal.

pSchlarb:
Hm I think we would have to create a `pytest.ini` since as i looked it isn't possible to set 2 working directories for the test discovery with the vscode settings.

.devcontainer/Dockerfile
pSchlarb: The container build also fails when trying to use the new version.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,31 @@
...
+		"mhutchie.git-graph",
+		"eamodio.gitlens",
+		"ms-python.python"
+	],
+
+	// Use 'forwardPorts' to make a list of ports inside the container available locally.
+	// "forwardPorts": [],
+
+	// Use 'postCreateCommand' to run commands after the container is created.
+	"postCreateCommand": "pip install .[tests]",
```
WadeBarnes, line 27:
```suggestion
	"postCreateCommand": "pip install flake8==3.8.4 .[tests]",
```

Install `flake8` otherwise there are Python (Output - Python) errors reported indicating `flake8` is not installed.

WadeBarnes (changes requested):
One small recommendation.

I'm also looking into some issues running the tests in the container.  Some tests "appear" to fail with a "Test result not found" error, when in fact they either pass or are skipped as expected.  You can see the true summary in the Output - Python Test Log console.  The issue appears to be a bug in the https://github.com/microsoft/vscode-python extension.  There have already been some fixes for related issues.

WadeBarnes:
> I'm also looking into some issues running the tests in the container. Some tests "appear" to fail with a "Test result not found" error, when in fact they either pass or are skipped as expected. You can see the true summary in the Output - Python Test Log console. The issue appears to be a bug in the https://github.com/microsoft/vscode-python extension. There have already been some fixes for related issues.

I've confirmed (through the Output - Python Test Log console) all of the tests pass or are skipped as expected.  Some of the `indy_node` tests fail if they are all run at once, which we know happens and is one of the reasons we split up the tests in the GHA workflow(s) as well.

The tests affected by the "Test result not found" issue are marked with a red circle with a red dot in the middle rather than a red circle with a red `x` inside.  All of the affected tests are parameterized tests, although in some cases there are some tests within the same group where the test results are found and others where the test results are not found.  For example:
image

WadeBarnes:
@pSchlarb, Could you also add some documentation to the top of the [Dev Setup](https://github.com/hyperledger/indy-node/blob/ubuntu-20.04-upgrade/docs/source/setup-dev.md) document about using the VSCode Devcontainer environment as the preferred method to create a development environment and get started with the code?

WadeBarnes:
The issues with the tests have been marked as a known issue here; https://github.com/hyperledger/indy-node/issues/1729

WadeBarnes:
Request for documentation has been added as a separate issue here; https://github.com/hyperledger/indy-node/issues/1728

WadeBarnes (approved):

Issue #1693, opened 2021-09-08
Multiple Dependency Installs in GHA


pSchlarb:
Noticed, that when slicing the modules for the tests each slice is installing the pip dependencies.
Would it be beneficial to include the dependency install into the Dockerfile? 
For example add the following pip command from `.github/workflows/build.yaml:232` to the `.github/workflows/build/Dockerfile`?
```      - name: Install dependencies
        run: |
          # Explicitly use the existing pip cache location in the node-build image.
          pip --cache-dir /root/.cache/pip install .[tests]
```

WadeBarnes:
I played with various ways of caching the dependency install when I was originally working on that part of the workflow.  The issue I ran into with moving it to the Dockerfile was the specific image then becomes the dependency and any given workflow run needs to use the image it built throughout the entire workflow, as at any time someone could check-in a new version of code that changes the installed dependencies.  The resulting image is only relevant to that code and that workflow, and may not work with previous or future workflow runs containing other code.  At the time I could not come up with a clean solution so went back to having the tests install the dependencies on their own.  The pip packages are cached, which helps to speed things up a small amount, but I've run into situations where even caching the pip packages causes problems with installing the dependencies.

I agree that there is a great deal of time waisted installing the dependencies before each test run, I just don't know the best way to address it at the moment.

swcurran:
Would there be a way to use a tag on the image, or even derive the image name (e.g., from the PR #) so that the test instances that need the one image could find it, and the next run would use a different name?

WadeBarnes:
The concern there is filling up the container registry with a bunch of onetime use container images.  I started going down that path at one point until I realized the ramifications (the mess getting left behind).

swcurran:
I was thinking that. Could there be an "end-of-tests" task that knows when all the tests have run and deletes the image from the registry?

WadeBarnes:
Perhaps.  I was also playing with generating the image internally and passing it around to the different jobs, but that did not work because of how the image is being used by the jobs.

pSchlarb:
As i understand the actual workflow, we are already creating and caching image for each run or am i missing something?
Would enabling Caching on the docker layers as described [here](https://evilmartians.com/chronicles/build-images-on-github-actions-with-docker-layer-caching) make sense?
Since the baseinstall with the apt-dependencies would not change often, compared to the pip dependency install?

WadeBarnes:
[...]

PR #1719, opened 2021-12-09
GHA Optimization

pSchlarb:
As discussed in #1693 this PR fixes the multiple installs in the test slices by moving the installation of the pip packages into the container build.
It als incorporates the #1716  in  the ubuntu20 branch.
As the container is special for a specific commit it is therefore tagged with the commit hash and deleted afterwards.
The script in the workflow folder is getting the package id from the container tag to delete it with api calls. 
The Linting can be moved outside of a container and therefore reducing the runtime.

Changes to remove_test_image job the need to be made to be used in the organiszation repo (the api calls need to be changed to fit the organisation not the user  part). 
But doing them now would fail the pipeline.
Will do that and sqash it after PR is approved.


.github/workflows/build/Dockerfile.ubuntu-2004
```diff
@@ -76,3 +76,7 @@ RUN gem install --no-document fpm -v 1.13.1
 
 RUN apt-get -y autoremove \
     && rm -rf /var/lib/apt/lists/*
+
+COPY . /indy-node
+WORKDIR /indy-node
+RUN pip install .[tests]
```
WadeBarnes, line 82: @pSchlarb, What is the impact of this on the `build_release` and `build_3rd_party_dependencies` jobs?  The build environment is not clean since it already has a code of the source code on it.

With the new `build-image` and `remove_test_image` process is there a possibility a `node-build` image might not exist by the time the `build_release` and `build_3rd_party_dependencies` jobs are run?

.github/workflows/getGHCRPackageID.py
```diff
@@ -0,0 +1,10 @@
+import json
+import sys
+
+sha = sys.argv[1]
+
+with open("apiResponse.json", "r") as api:
+    jsonApi = json.loads(api.read())
+    for i in jsonApi:
+        if sha in i['metadata']['container']['tags']:
+            print(i['id'])
```
WadeBarnes, line 10: Nit-picking ...

```suggestion
with open("apiResponse.json", "r") as apiResponse:
    data = json.loads(apiResponse.read())
    for i in data:
        if sha in i['metadata']['container']['tags']:
            print(i['id'])
```

.github/workflows/build.yaml
```diff
@@ -263,30 +180,73 @@ jobs:
...
+          PACKAGE_VERSION: ${{ steps.pversion.outputs.packageVersion }}
+        run: |
+            curl -u pschlarb:${{ secrets.API }} -H "Accept: application/vnd.github.v3+json" -X DELETE $GITHUB_API_URL/user/packages/container/node-build/versions/$PACKAGE_VERSION | grep "last tagged"
+            if [ $? -eq 0 ];
+            then 
+              echo $(curl -u pschlarb:${{ secrets.API }} -X DELETE -H"Accept: applicatiov/vnd.github.vs3+json" $GITHUB_API_URL/user/packages/container/node-build);
+            else 
+              echo done;
+            fi;
+
```
WadeBarnes, line 227: Is there no way to get this process to dynamically work on different repositories rather than hardcoding the username, or is that what the `GITHUB_REPOSITORY_OWNER` is going to be used for?

.github/workflows/build.yaml
```diff
@@ -35,25 +39,20 @@ jobs:
...
-            echo "::set-output name=CACHE_KEY_BUILD::${{ hashFiles('.github/workflows/build/Dockerfile.ubuntu-2004') }}"
-            echo "::set-output name=CACHE_KEY_LINT::${{ hashFiles('.github/workflows/lint/Dockerfile.ubuntu-2004') }}"
-            echo "::set-output name=UBUNTU_VERSION::ubuntu-2004"
-            echo "::set-output name=distribution::focal"
-          fi
-     
-
+          echo "::set-output name=UBUNTU_VERSION::ubuntu-2004"
+          echo "::set-output name=distribution::focal"
+          echo "::set-output name=shortSHA::sha-$(git rev-parse --short HEAD)"
```
WadeBarnes, line 54: @pSchlarb, @udosson Have we made a conscious decision to do away with the support for ubuntu 16.04 builds on this branch?  Would the plan then be to only support 16.04 builds on the main branch and 20.04 on this branch until such time as this branch is merged?  If so we should do away with all of the other 16.04 related CI/CD sources on this branch.

.github/workflows/build.yaml
```diff
@@ -93,123 +93,61 @@ jobs:
...
             org.opencontainers.image.created=${{ steps.prep.outputs.created }}
             org.opencontainers.image.revision=${{ github.sha }}
-
-      - name: Touch Cache
-        if: steps.cache-image.outputs.cache-hit != 'true'
-        run: |
-          mkdir -p ${GITHUB_WORKSPACE}/cache
-          touch ${GITHUB_WORKSPACE}/cache/${{ env.CACHE_KEY_BUILD }}
-
-  lint-image:
```
WadeBarnes, line 152: If the lint image is not needed anymore the related CI/CD sources and artifacts should be deleted.

WadeBarnes (commented):
Apart from my other comments I'm wondering whether we should still have a clean cached build image that is used for the builds and as the base for a separate test image that is only used for the tests and is managed like the build image is in this PR.

.github/workflows/build.yaml
```diff
@@ -263,30 +180,73 @@ jobs:
...
+    needs: [workflow-setup, indy_node_tests]
+    if: always()
+    runs-on : ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Run a one-line script
+        run: |
+            curl -u pschlarb:${{ secrets.API }} -H "Accept: application/vnd.github.v3+json" \
```
ryjones, line 205: this probably shouldn't be hardwired to pschlarb

.github/workflows/build.yaml
```diff
@@ -263,30 +180,73 @@ jobs:
...
+        id: pversion
+        env:
+          SHORT_SHA: ${{ needs.workflow-setup.outputs.shortSHA }}
+        run: |
+          echo "::set-output name=packageVersion::$(python .github/workflows/getGHCRPackageID.py $SHORT_SHA)"
+      - name: Remove Docker Container or Package with sha
+        env: 
+          PACKAGE_VERSION: ${{ steps.pversion.outputs.packageVersion }}
+        run: |
+            curl -u pschlarb:${{ secrets.API }} -H "Accept: application/vnd.github.v3+json" -X DELETE $GITHUB_API_URL/user/packages/container/node-build/versions/$PACKAGE_VERSION | grep "last tagged"
```
ryjones, line 220: same here

.github/workflows/build.yaml
```diff
@@ -263,30 +180,73 @@ jobs:
...
+            if [ $? -eq 0 ];
+            then 
+              echo $(curl -u pschlarb:${{ secrets.API }} -X DELETE -H"Accept: applicatiov/vnd.github.vs3+json" $GITHUB_API_URL/user/packages/container/node-build);
```
ryjones, line 223: and here

.github/workflows/build.yaml
```diff
@@ -21,11 +23,13 @@ jobs:
       # as the push and pull operations require the URL of the repository
       # to be in lowercase.
       GITHUB_REPOSITORY_NAME: ${{ steps.repository-name.outputs.lowercase }}
+      GITHUB_REPOSITORY_OWNER: ${{ steps.repository-owner.outputs.lowercase }}
```
WadeBarnes, line 26: Also, why the transition to using `github.repository_owner` rather than `github.repository` especially with the container repository URLs?

ryjones (changes requested):
Hi,
please take a look at [how Firefly does parts of this](https://github.com/hyperledger/firefly/blob/main/.github/workflows/docker_main.yml) [and use those patterns](https://github.com/hyperledger/firefly/blob/main/.github/workflows/docker_release.yml)

.github/workflows/build.yaml
pSchlarb: It was neccessary, because with `github.repository` the container would have another level ( [USER|ORG]/[repo]/node-build ), which couldn't be accessed by the package api.
When moved to the toplevel ( [USER|Org]/node-build) it was. 
Not sure if that is problematic. 
Haven't checked if the dynamic removal of the images could be achieved by using the DockerApi of GHCR and preserving the `github.repository` for the container urls.

.github/workflows/build.yaml
pSchlarb: > It was neccessary because with `github.repository` the container would have another level ( [USER|ORG]/[repo]/node-build ), which couldn't be accessed by the package api.

And yes it is going to be used for dynamically use of it.

.github/workflows/build/Dockerfile.ubuntu-2004
pSchlarb: Good  catch totally overlooked these.
I think there is no impact on them, as each Job starts its own container and they would run in the container for the actual commit. 
Will change them to use the code already present in the container.

pSchlarb:
@ryjones Totally agree with the hardcoding.
Your suggestion to use the `github_token` for deletion can't be used as it can't have the `delete:package` permission (see [here](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token)). 
Could we get around the hardcodung by using the `github.repository_owner and an accesstoken as a secret?
> Ah, the issue is that PRs don't have access to secrets.
Is it possible to resolve that by accessing the secrets via environment variables as described [here](https://docs.github.com/en/actions/security-guides/encrypted-secrets) 
Or would that be too insecure?

.github/workflows/build.yaml
pSchlarb: Would be in favor for that.

ryjones:
@pSchlarb I think the easiest way to do this is on a branch on this repo, instead of a PR from another repo. So a branch of `hyperledger:ubuntu-20.04-upgrade`

pSchlarb:
@ryjones I tried to create a branch, but it appears to me that i don't have the required permissions.

ryjones:
@pSchlarb [here you go](https://github.com/hyperledger/indy-node/tree/gha_optimizing).

ryjones:
@pSchlarb I updated [the branch here](https://github.com/hyperledger/indy-node/tree/gha_optimizing) to match your PR. You should have permissions to directly work on that branch now, which should enable all of the actions.

pSchlarb:
Will close this PR in favor for #1721

PR #1687, opened 2021-08-18
Added ursa to the core.ubuntu.dockerfile install

pSchlarb:
Testpool started with the pool_start.sh script after the changes
Fixes #1624

Signed-off-by: Philipp Schlarb <p.schlarb@esatus.com>

PR #1724, opened 2021-12-31
Fix indy-plenum version reference in install_requires

WadeBarnes:
- The convention was changed inadvertently to a `.` rather than a `~`.  The overwhelming majority of `dev` releases use the `~`.

Signed-off-by: Wade Barnes <wade@neoterictech.ca>

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -54,7 +54,7 @@ setup(
     data_files=[(
         (BASE_DIR, ['data/nssm_original.exe'])
     )],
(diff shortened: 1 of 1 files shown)


WadeBarnes:
This has caused a few issues like this one; https://github.com/hyperledger/indy-node/pull/1687#issuecomment-1003435213

WadeBarnes:
This is not the problem.

PR #1725, opened 2022-01-04
Indy-Test-Autmation

udosson:
This PR contains the workflow to run the `indy-node-tests` from https://github.com/hyperledger/indy-test-automation.

Currently, the workflow is only triggered if the `GITHUB_REF` is `rc`.
The trigger of the `indy-test-automation` workflow needs some more automation in terms of setting the input parameters dynamically.

The `WORKFLOW_DISPATCH_TOKEN` that is used to trigger the workflow is not accessible to PRs. (https://chat.hyperledger.org/channel/indy-contributors?msg=Gxj3huBNEMyCNiNwP)

A successful run of the `indy-test-automation` workflow is at https://github.com/udosson/indy-node/actions/runs/1687056439

The workflow currently only works for Indy based on Ubuntu 16.04 because a suitable `systemd` image for the Ubuntu 20.04 Node container is still missing.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -275,6 +275,8 @@ jobs:
     runs-on: ubuntu-20.04
     container:
       image: ghcr.io/${{ needs.workflow-setup.outputs.GITHUB_REPOSITORY_NAME }}/node-build:ubuntu-16-04
+    outputs:
+      INDY_NODE_PACKAGE_VERSION: ${{ steps.cache.outputs.pkgVer }}
     steps:
       - name: Check out code
         uses: actions/checkout@v1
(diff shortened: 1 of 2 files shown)


udosson:
(ci) test this please

.github/workflows/indy_test_automation.yml
```diff
@@ -0,0 +1,1677 @@
...
+          echo ::set-output name=tags::${TAGS}
+          echo ::set-output name=created::$(date -u +'%Y-%m-%dT%H:%M:%SZ')
+
+      - name: Log into the GitHub Container Registry
+        if: steps.cache-image.outputs.cache-hit != 'true'
+        uses: docker/login-action@v1
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
```
WadeBarnes, line 170: For the node and client image builds, is it necessary to log into the registry when the images are just being cached and uploaded as artifacts to the following steps?

.github/workflows/build.yaml
```diff
@@ -275,6 +275,8 @@ jobs:
     runs-on: ubuntu-20.04
     container:
       image: ghcr.io/${{ needs.workflow-setup.outputs.GITHUB_REPOSITORY_NAME }}/node-build:ubuntu-16-04
+    outputs:
+      INDY_NODE_PACKAGE_VERSION: ${{ steps.cache.outputs.pkgVer }}:
```
WadeBarnes, line 279: This line is causing and error and stopping the workflow from being run.

image

WadeBarnes (changes requested):
Could the tests be setup in a matrix to reduce the code duplication?

.github/workflows/build.yaml
udosson: I'll fix it. The colon at the end of the line causes the error

.github/workflows/build.yaml
```diff
@@ -422,4 +424,40 @@ jobs:
...
+          libIndyVersion=$(grep -oP "(?<=python3-indy==).*?(?=')" <<< "$(cat setup.py)")-xenial
+          echo "::set-output name=libIndyVersion::${libIndyVersion}"
+       
+      - name: Invoke workflow with inputs
+        uses: benc-uk/workflow-dispatch@v1
+        with:
+          workflow: indy-test-automation
+          token: ${{ secrets.WORKFLOW_DISPATCH_TOKEN }}
+          inputs: '{ "nodeVersion": "1.13.0~dev197", "plenumVersion": "${{ steps.version.outputs.plenumVersion }}", "ursaVersion": "0.3.2-2", "pyzmqVersion": "18.1.0", "libIndyVersion": "1.15.0~1625-xenial", "ubuntuVersion": "ubuntu-1604", "nodeRepoComponent": "main", "nodeSovrinRepoComponent": "master",  "clientSovrinRepoComponent": "master" }'
+         # inputs: '{ "nodeVersion": "${{ env.INDY_NODE_PACKAGE_VERSION }}", "plenumVersion": {{ steps.version.outputs.plenumVersion }}, "ursaVersion": "0.3.2-2", "pyzmqVersion": "18.1.0", "libIndyVersion": {{ steps.version.outputs.libIndyVersion }},, "ubuntuVersion": "ubuntu-1604", "nodeRepoComponent": "main", "nodeSovrinRepoComponent": "master",  "clientSovrinRepoComponent": "master" }'
```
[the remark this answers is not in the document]
udosson: Just temporarily. I haven't had the time to set the variables dynamically.
We could open an issue for this. What do you think?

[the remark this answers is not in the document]
WadeBarnes: I think an issue would be good, other wise we might forget.

.github/workflows/indy_test_automation.yml
```diff
@@ -0,0 +1,1672 @@
...
+          retention-days: 1
+      
+      - name: Move cache
+        run: |
+          rm -rf /tmp/.buildx-cache
+          mv /tmp/.buildx-cache-new /tmp/.buildx-cache                     
+
+  ### 1 step in workflow  per test
+  ### The tests rely on docker in docker with the fixed network name and fixed IP addresses. 
+  ### That's why the tests cannot be run in matrix mode because all tests would share the same host and some docker engine.
```
WadeBarnes, line 281:
```suggestion
  ### That's why the tests cannot be run in matrix mode because all tests would share the same host and same docker engine.
```

WadeBarnes (approved):

PR #1748, opened 2022-05-06
fix importlib-metadata dependency

mgmgwi:
This should fix the wrong dependency in the indy-node .deb package.

> The following packages have unmet dependencies:
>  indy-node : Depends: importlib-metadata (< 3.0) but it is not installable

Signed-off-by: mgmgwi <guido.wischrop@mgm-tp.com>

### DIFF
--- a/build-scripts/ubuntu-2004/prepare-package.sh
+++ b/build-scripts/ubuntu-2004/prepare-package.sh
@@ -28,12 +28,14 @@ echo -e "\n\nPrepares indy-node debian package version"
 sed -i -r "s~indy-node==([0-9\.]+[0-9])(\.)?([a-z]+)~indy-node==\1\~\3~" setup.py
 
 if [ "$distro_packages" = "debian-packages" ]; then
-  # Only used for the deb package builds, NOT for the PyPi package builds.
-  # Update the package names to match the versions that are pre-installed on the os.
+  # Update the package names to match the names that are available on the os.
   echo -e "\nAdapt the dependencies for the Canonical archive"
   sed -i "s~timeout-decorator~python3-timeout-decorator~" setup.py
   sed -i "s~distro~python3-distro~" setup.py
+  sed -i "s~importlib-metadata=~python3-importlib-metadata=~" setup.py
 
+
+  # Only used for the deb package builds, NOT for the PyPi package builds.
   echo -e "\n\nPrepares indy-plenum debian package version"
   sed -i -r "s~indy-plenum==([0-9\.]+[0-9])(\.)?([a-z]+)~indy-plenum==\1\~\3~" setup.py
 
--- a/setup.py
+++ b/setup.py
@@ -56,7 +56,11 @@ setup(
     )],
 
     install_requires=['indy-plenum==1.13.0.dev14',
-                      'importlib-metadata<3.0',
+                    # importlib-metadata needs to be pinned to 3.10.1 because from v4.0.0 the package
+                    # name ends in python3-importlib-metadata_0.0.0_amd64.deb
+                    # see also build-scripts/ubuntu-2004/build-3rd-parties.sh
+                    # https://github.com/hyperledger/indy-plenum/blob/eac38674252b539216be2c40bb13e53c5b70dad2/build-scripts/ubuntu-2004/build-3rd-parties.sh#L104-L106
+                      'importlib-metadata==3.10.1',
                       'timeout-decorator>=0.5.0',
                       'distro>=1.5.0'],
     setup_requires=['pytest-runner'],


WadeBarnes:
Thanks @mgmgwi.  @udosson, @pSchlarb, this PR is relying on the fact that `indy-plenum` is building and publishing this particular 3rd party package; https://github.com/hyperledger/indy-plenum/blob/ubuntu-20.04-upgrade/build-scripts/ubuntu-2004/build-3rd-parties.sh#L106.  How do we feel about this implicit dependency?  Perhaps a comment in the prepare package script would be enough.

pSchlarb:
Hm since version setting is also commented in plenum, i see no problem with that so far.
IMHO a comment should be sufficient and wouldn't hurt.

build-scripts/ubuntu-2004/prepare-package.sh
```diff
@@ -33,6 +33,7 @@ if [ "$distro_packages" = "debian-packages" ]; then
   echo -e "\nAdapt the dependencies for the Canonical archive"
   sed -i "s~timeout-decorator~python3-timeout-decorator~" setup.py
   sed -i "s~distro~python3-distro~" setup.py
+  sed -i "s~importlib-metadata~python3-importlib-metadata~" setup.py
```
udosson, line 36: @mgmgwi could you remove this line? 
We don't use `python3-importlib-metadata` from the Canonica archive because `python3-importlib-metadata` is still at version `1.5.0-1` there (https://packages.ubuntu.com/focal/python3-importlib-metadata)

We build the Debian package based on the PyPi packages of `python3-importlib-metadata`, publish it to Hyperledger's Artifactory and pull the artifact from there.

udosson (changes requested):
@mgmgwi thanks for the PR and spotting the issue.
Could you please make the small change?

udosson:
I'd rather add the comment to `setup.py` because we don't need to change the `prepare-package.sh` script

setup.py
```diff
@@ -56,7 +56,7 @@
     )],
 
     install_requires=['indy-plenum==1.13.0.dev14',
-                      'importlib-metadata<3.0',
+                      'importlib-metadata==3.10.1',
```
udosson, line 59: Could you add a comment here that says something like:
"Needs to be pinned to because of fpm build issues with a higher version number?
@WadeBarnes @pSchlarb do you have a better suggestion?

pSchlarb: Not really. Maybe also a cross reference to https://github.com/hyperledger/indy-plenum/blob/ubuntu-20.04-upgrade/build-scripts/ubuntu-2004/build-3rd-parties.sh#L104-L106 .

build-scripts/ubuntu-2004/prepare-package.sh
mgmgwi: Hmm, the problem is, that the .deb package contains a reference to `importlib-metadata` which we have to replace or remove. If not here, where then?

udosson: The dependencies are taken from `setup.py`. `importlib-metadata==3.10.1` is referenced there already. 
When installing the Debian package of Indy-Node `dpgk` looks for `python3-importlib-metadata==3.10.1` which will be pulled from here https://hyperledger.jfrog.io/ui/native/indy/pool/focal/dev/i/importlib-metadata/

mgmgwi: :thinking: are you sure? I think it will look for `importlib-metadata=3.10.1` (missing the `python3-` prefix) which still does not exists. I tried manually patching the control file in the .deb package like this:

```
Package: indy-node
Version: 1.13.0~dev24
License: Apache 2.0
Vendor: none
Architecture: amd64
Maintainer: Hyperledger <hyperledger-indy@lists.hyperledger.org>
Installed-Size: 2807
Depends: indy-plenum (= 1.13.0~dev14), importlib-metadata (= 3.10.1), python3-timeout-decorator (>= 0.5.0), python3-distro (>= 1.5.0), at, iptables, libsodium23
Section: default
Priority: extra
Homepage: https://github.com/hyperledger/indy-node
Description: Indy node
```
Please note the `Depends` line. Then I repackaged it, but when I install it with `apt install /tmp/indy-node_1.13.0_dev24-patched2_amd64.deb` it still complains:

```
The following packages have unmet dependencies:
 indy-node : Depends: importlib-metadata (= 3.10.1) but it is not installable
```

setup.py
mgmgwi: Done.

build-scripts/ubuntu-2004/prepare-package.sh
udosson: I'm quite sure but I cannot verify before Tuesday. 
We use `fpm` to build the Debian artifacts with the following flag `--python-package-name-prefix "python3"`
All `install_requires` (https://github.com/hyperledger/indy-node/blob/ubuntu-20.04-upgrade/setup.py#L58) packages will be converted to `python3-package_name`. 
Therefore we publish the third-party dependencies with the `python3` prefix. And in case we use the artifacts from the canonical archive we adjust them to the correct build version. An example for Indy Plenum can be found [here ](https://github.com/hyperledger/indy-plenum/blob/master/build-scripts/ubuntu-1604/prepare-package.sh#L31) for Ubuntu 26.04

mgmgwi:
I removed the sed replacement in [build-scripts/ubuntu-2004/prepare-package.sh#L34](https://github.com/hyperledger/indy-node/blob/a0dc19dc77445e35cf40aa512c40f34b4d6af3dd/build-scripts/ubuntu-2004/prepare-package.sh#L34) as requested by @udosson in the comment above.

udosson:
@mgmgwi - thank you for the changes. I'll test the build and install process on Tuesday and will merge after the successful testing.

udosson:
Thanks for adjusting the PR according to our discussion, @mgmgwi 

We need to at the `python3-` prefix in `prepare-package.sh`. However, I cannot say at the moment why we don't need this in Indy Plenum. I'll do further investigations and please let me know your thoughts @WadeBarnes and @pSchlarb if you have an idea.

udosson (approved):
LGMT. Thanks, @mgmgwi

WadeBarnes:
> Thanks for adjusting the PR according to our discussion, @mgmgwi
> 
> We need to at the `python3-` prefix in `prepare-package.sh`. However, I cannot say at the moment why we don't need this in Indy Plenum. I'll do further investigations and please let me know your thoughts @WadeBarnes and @pSchlarb if you have an idea.

Perhaps an oversight.

PR #1762, opened 2022-06-06
[v1.13.1-rc2] - Update Version Number for Release

sovbot:
[v1.13.1-rc2] - Update Version number for Release


WadeBarnes:
Uncovered an issue with the workflows.  Needed to close before the version was set.  Will run another release cycle once the fix is merged.

PR #1794, opened 2023-01-19
Fix dependency scheme

WadeBarnes:
- A fix for the dependency scheme related to indy-plenum.  Fixes the issue where the python package version installed using the deb package is different than the published python package.

- The issue was due to how the dependency options of `fpm` were being used.  The deb package build was using the python dependencies to define the deb package dependencies.  Since the deb and python package version numbers use slightly different formats for non-release (`dev` and `rc`) versions, the versions of the python dependencies (notably indy-plenum) were being set to match the deb version format in order to correctly define the deb dependencies.  This strategy has the unfortunate side affect of causing dependency issues with the python packages down the road.

Signed-off-by: Wade Barnes <wade@neoterictech.ca>

### DIFF
--- a/.github/workflows/tag.yaml
+++ b/.github/workflows/tag.yaml
@@ -55,7 +55,7 @@ jobs:
                        libgflags-dev
           pip install packaging \
                       importlib_metadata==3.10.1 \
-                      indy-plenum==1.13.1rc3 \
+                      indy-plenum==1.13.1.rc3 \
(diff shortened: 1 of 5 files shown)


WadeBarnes:
A copy of the resulting deb package will be attached to this PRs workflow run when it completes.

pSchlarb (approved):

PR #1812, opened 2023-04-18
Repo Dispatches and dispatchable WF-Runs

pSchlarb:
Automatically creates PR when node is triggered by plenum and triggers an update to token-plugin and sovrin, when a release is created.

Needs the `BOT_PR_PAT_NODE` secret set with a fine grained access token for sovrin /token-plugin so that it can trigger the workflows there.


WadeBarnes (changes requested):
Consider creating a shared action for the `Convert to python version flavour` step.  The same process is duplicated here and in the other related PRs.

The workflows still contain references to your repos.

.github/workflows/publishRelease.yaml
```diff
@@ -119,3 +119,41 @@ jobs:
...
+
+  triggerTokenUpdate:
+    runs-on: ubuntu-latest
+    needs: [release-infos, publish_artifacts, convertPyVersion]
+    steps:
+      - name: Repository Dispatch
+        uses: peter-evans/repository-dispatch@v2
+        with:
+          token: ${{ secrets.BOT_PR_PAT }}
+          repository: sovrin-foundation/token-plugin-1
```
WadeBarnes, line 157:
```suggestion
          repository: sovrin-foundation/token-plugin
```

WadeBarnes (changes requested):

.github/workflows/publishRelease.yaml
WadeBarnes: Still an issue

WadeBarnes (changes requested):

WadeBarnes (approved):

PR #1829, opened 2024-02-01
RTD Fixes to enable publishing

swcurran:
Signed-off-by: Stephen Curran <swcurran@gmail.com>


swcurran:
An update that should re-enable updating the RTD site.  There are some warnings -- missing documents and the like -- that we can clean up after we see if this will fix the RTD generation errors.

WadeBarnes (approved):

WadeBarnes:
Since the build issues have nothing to do with these changes, we'll deal with those separately; https://github.com/hyperledger/indy-node/issues/1830

PR #1898, opened 2025-01-15
Adds OpenSSF scorecard workflow and badge

swcurran:
Addresses #1897  by adding the badge and the scorecard task that will execute weekly.  Once this is merged, I'll run it manually. Until it has been run at least once. the badge won't generate and the link won't resolve. But that's OK -- we'll get it run.


swcurran:
Pretty sure the failure is unrelated to the PR, and we can go ahead and merge (with an approval).

lynnbendixsen (approved):

PR #1900, opened 2025-01-24
Update logos

ryjones:


swcurran (dismissed):
Should the png logos be sized the same?  I’m not a designer, but I would think depending on where the logos are being used, the logo size difference would matter.

That said, approving, so you can go ahead if this is not an issue.

WadeBarnes (approved):

ryjones:
I can do another round

ryjones:
OK, I made the logo the same height, but it is a little wider.

swcurran (approved):
Thanks — LGTM

PR #1904, opened 2025-02-07
Build and workflow updates and fixes

WadeBarnes:
- Fix Plenum version format in the indy-node deb package.
- Enable workflows on the `ubuntu-22.04` branch.
- Set concurrency rules on PR and Push workflows.
  - Cancel in progress runs when stating a new one.
- Ensure tests are running on latest version of Ubuntu, and with Python 3.10 (the default on Ubuntu 22.04).


swcurran (approved):
Is the use of “latest” make bring in 24.04 possible?  I’m assuming not, and the change was intentional — perhaps it being “latest” within a release (20.04, 22.04, etc.).

Approving to not holding things up.

WadeBarnes:
> Is the use of “latest” make bring in 24.04 possible? I’m assuming not, and the change was intentional — perhaps it being “latest” within a release (20.04, 22.04, etc.).

I did that on purpose to see the results.  I was initially going to specify 22.04.

GuillaumeBourque-QC:
> Is the use of “latest” make bring in 24.04 possible? I’m assuming not, and the change was intentional — perhaps it being “latest” within a release (20.04, 22.04, etc.).
> 
> Approving to not holding things up.

If we look here: https://github.com/actions/runner-images  we see that ubuntu-latest is already using 24.04

image

WadeBarnes:
> > Is the use of “latest” make bring in 24.04 possible? I’m assuming not, and the change was intentional — perhaps it being “latest” within a release (20.04, 22.04, etc.).
> > Approving to not holding things up.
> 
> If we look here: https://github.com/actions/runner-images we see that ubuntu-latest is already using 24.04

I am aware.  I looked this up before making the changes.  I want to see the affect of using 24.04 as the runner.

swcurran (approved):
LGTM — I’ll leave it to you to merge.

swcurran (approved):

WadeBarnes:
Ok, the plenum reference is fixed.  But a couple other packages can't be found.  So we need to look at those next.

WadeBarnes:
> Since in plenum the version is [1, 14, 0, "rc", 0] should we use the same version in indy-node ?

That will come in time when we trigger an `rc` release.

WadeBarnes:
The two package that can't be found, `python3-timeout-decorator 0.5.0`, and `python3-distro 1.7.0`, are built during the build 3rd party package stage of the build.  Since PRs don't have the permissions to publish the artifacts, they don't get published and therefore can't be found when `indy-node` is being installed by the test runners.  I'll need to publish those manually.

GuillaumeBourque-QC:
But these should be from the OS ?

if we search we get these:

> apt search python3-timeout-decorator
...

> python3-timeout-decorator/jammy 0.5.0-1 all

The same is true for the second one.

> python3-distro/jammy,now 1.7.0-1 all [installed,automatic]

Are we just missing an apt update before the package install ? Or do we want to provide a copy of those pakages ?

GuillaumeBourque-QC:
if we specify this version apt install  python3-timeout-decorator=0.5.0-1 it will work anbd the same is true for the other one.

So maybe something in the regex is cutting the last part of the debian package name

WadeBarnes:
> But these should be from the OS ?
> 
> if we search we get these:
> 
> > apt search python3-timeout-decorator
> > ...
> 
> > python3-timeout-decorator/jammy 0.5.0-1 all
> 
> The same is true for the second one.
> 
> > python3-distro/jammy,now 1.7.0-1 all [installed,automatic]
> 
> Are we just missing an apt update before the package install ? Or do we want to provide a copy of those pakages ?

The `fpm` package build in the [build-indy_node.sh](https://github.com/hyperledger/indy-node/blob/ubuntu-22.04/build-scripts/ubuntu-2204/build-indy_node.sh) build script extracts the dependencies (by default) from the indy-node python dependencies.  The [build-3rd-parties.sh](https://github.com/hyperledger/indy-node/blob/ubuntu-22.04/build-scripts/ubuntu-2204/build-3rd-parties.sh) script builds the packages to match those dependencies.  The [prepare-package.sh](https://github.com/hyperledger/indy-node/blob/ubuntu-22.04/build-scripts/ubuntu-2204/prepare-package.sh) script, does some alignment between available package(s) from the OS and the listed indy-node dependencies.

So we might just be able to adapt the dependencies in the [prepare-package.sh](https://github.com/hyperledger/indy-node/blob/ubuntu-22.04/build-scripts/ubuntu-2204/prepare-package.sh) script.

The indy-plenum [prepare-package.sh](https://github.com/hyperledger/indy-plenum/blob/ubuntu-22.04/build-scripts/ubuntu-2204/prepare-package.sh) script provides some better examples of how this is done.

I'll try fixing the dependencies this way.

swcurran:
Merged with failing tests because this will not be the last PR — allows others to work on the latest changes to fix more things.

Issue #1908, opened 2025-02-11
Remove obsolete `dev-setup` documentation


WadeBarnes:
The [dev-setup](https://github.com/hyperledger/indy-node/tree/main/dev-setup) folder contains obsolete documentation on how to setup a local development environment.

The requirement for bare metal setup has been eliminated by implementing devContainers, which provide a repeatable and sharable stable development environment.

References to any of the bare metal setup should be replaced with references to the devContainer utilization documentation.

WadeBarnes:
@IshaanXCoder, You don't need to be assigned the task to work on it.  An indication here that you are, plus a relation to an associated PR is all that is needed.  In other words - Go for it!

bougui:
Hello here is a PR for this https://github.com/hyperledger/indy-node/pull/1914

PR #1914, opened 2025-02-19
Dev setup precision to use devcontainer

bougui:
In reference to this request Remove obsolete dev-setup documentation #1908


WadeBarnes:
I'd recommend getting rid of the whole `dev-setup` folder.

bougui:
@WadeBarnes,  done like proposed.

Guillaume

WadeBarnes (changes requested):
Great.  However the `./docs/source/setup-dev.md`, which is referenced on the main page, still contains references to the `dev-setup` folder as well as instructions on how to set up a bare metal development environment manually.  That document could be deleted and the relevant content about using the dev container moved out to the `How to Start Working with the Code` section of the main README.

WadeBarnes:
> @WadeBarnes I would keep the file since the python instructions are still relevant and it make the main README.md file bigger If I bring back what was in the setup-dev.md. Since we already have a seperate file I propose to only delete part refering to the code that is not maintain anymore.

If there are still manual development environment setup steps that are relevant, I'd suggest those steps be incorporated into the dev container setup so everyone has a consistent development environment.  However in this case I'd argue the python virtual environment is more relevant to a bare metal setup where the user is more likely to have a conflicting version of python installed.  The dev container sets up the version of python supported by the corresponding version of Ubuntu and code, and it's already a virtual environment.

WadeBarnes (approved):

